PR #6, opened 2017-07-24
Change to re-create masterkeys

sharidas:
This change brings a new command to re-create
masterkey

### DIFF
--- a/lib/Command/RecreateMasterKey.php
+++ b/lib/Command/RecreateMasterKey.php
@@ -1,0 +1,343 @@
+<?php
+/**
+* @author Sujith Haridasan <sharidasan@owncloud.com>
+*
+* @copyright Copyright (c) 2017, ownCloud GmbH
+* @license AGPL-3.0
+*
+* This code is free software: you can redistribute it and/or modify
+* it under the terms of the GNU Affero General Public License, version 3,
+* as published by the Free Software Foundation.
+*
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+* GNU Affero General Public License for more details.
+*
+* You should have received a copy of the GNU Affero General Public License, version 3,
+* along with this program.  If not, see <http://www.gnu.org/licenses/>
+*
+*/
+
+
+namespace OCA\Encryption\Command;
+
+use OC\Encryption\CustomEncryptionWrapper;
+use OC\Encryption\Exceptions\DecryptionFailedException;
+use OC\Encryption\Manager;
+use OC\Files\CustomView;
+use OC\Files\Filesystem;
+use OC\Files\View;
+use OC\Memcache\ArrayCache;
+use OCA\Encryption\Crypto\EncryptAll;
+use OCA\Encryption\KeyManager;
+use OCA\Encryption\Users\Setup;
+use OCA\Encryption\Util;
+use OCP\App\IAppManager;
+use OCP\IAppConfig;
+use OCP\IConfig;
+use OCP\IL10N;
+use OCP\ILogger;
+use OCP\ISession;
+use OCP\IUserManager;
+use OCP\Mail\IMailer;
+use OCP\Security\ISecureRandom;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Helper\ProgressBar;
+use Symfony\Component\Console\Helper\QuestionHelper;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
+
+class RecreateMasterKey extends Command {
+
+	/** @var Manager  */
+	protected $encryptionManager;
+
+	/** @var IUserManager  */
+	protected $userManager;
+
+	/** @var View  */
+	protected $rootView;
+
+	/** @var KeyManager  */
+	protected $keyManager;
+
+	/** @var Util  */
+	protected $util;
+
+	/** @var  IAppManager */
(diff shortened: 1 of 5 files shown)


lib/Command/RecreateMasterKey.php
```diff
@@ -0,0 +1,320 @@
...
+		parent::configure();
+
+		$this
+			->setName('encryption:recreate-master-key')
+			->setDescription('Replace existing master key with new one. Encrypt the file system with newly created master key')
+		;
+	}
+
+	protected function execute(InputInterface $input, OutputInterface $output) {
+		if ($this->util->isMasterKeyEnabled()) {
```
PVince81: I think we should add a warning here before starting and asking whether to proceed "Y/N". Please check how warnings are displayed for the other encryption commands. Also add a "-y" option to answer yes automatically.

Considering that this will decrypt a whole instance and might take days, we better inform the admin in advance before starting.

cc @tomneedham

lib/Command/RecreateMasterKey.php
```diff
@@ -0,0 +1,320 @@
...
+			$this->keyManager->validateShareKey();
+			$this->keyManager->validateMasterKey();
+			$progress->setMessage("Encryption progress...");
+			$progress->start();
+			$progress->advance();
+			$this->encryptAllUsersFiles($progress);
+			$progress->finish();
+			$output->writeln("\nEncryption completed successfully\n");
+		} else {
+			$output->writeln("Master key is not enabled. Kindly enable it\n");
```
PVince81: please remove the "Kindly" bit. The instance might be in "user key" mode

lib/Command/RecreateMasterKey.php
```diff
@@ -0,0 +1,320 @@
...
+		}
+	}
+
+	public function encryptAllUsersFiles(ProgressBar $progress) {
+		$userNo = 1;
```
PVince81: start at zero ?

lib/Command/RecreateMasterKey.php
```diff
@@ -0,0 +1,320 @@
...
+						$progress->advance();
+					}
+					$userNo++;
+				}
+				$offset += $limit;
+			} while(count($users) >= $limit);
+		}
+	}
+
+	public function encryptUsersFiles($uid, ProgressBar $progress) {
```
PVince81: is there any code from the occ encrypt-all command we could reuse ?

lib/Command/RecreateMasterKey.php
```diff
@@ -0,0 +1,320 @@
...
+	protected $util;
+
+	/** @var  IAppManager */
+	protected $IAppManager;
+
+	/** @var  IAppConfig */
+	protected $appConfig;
+
+	/** @var IConfig  */
+	protected $IConfig;
```
PVince81: local attributes must start with a lower case and be camel case.

lib/Command/RecreateMasterKey.php
```diff
@@ -0,0 +1,320 @@
...
+							$this->failed[$uid][] = $path;
+						} else {
+							$this->failed[$uid] = [$path];
+						}
+					}
+				}
+			}
+		}
+
+		if (empty($this->failed)) {
```
PVince81: show error in case of failure ? an admin might want to ctrl+c when seeing this in real-time

tests/Command/RecreateMasterKeyTest.php
```diff
@@ -0,0 +1,302 @@
...
+	protected $util;
+
+	/** @var  IAppManager | \PHPUnit_Framework_MockObject_MockObject */
+	protected $IAppManager;
+
+	/** @var  IAppConfig | \PHPUnit_Framework_MockObject_MockObject */
+	protected $appConfig;
+
+	/** @var IConfig | \PHPUnit_Framework_MockObject_MockObject */
+	protected $IConfig;
```
PVince81: see my comment on attribute casing above

lib/Command/RecreateMasterKey.php
```diff
@@ -0,0 +1,318 @@
...
+		}
+	}
+
+	protected function encryptAllUsers(InputInterface $input, OutputInterface $output) {
+		$this->encryptAll = new EncryptAll(
+			$this->userSetup, $this->userManager, $this->rootView,
+			$this->keyManager, $this->util, $this->config,
+			$this->mailer, $this->l, $this->questionHelper,
+			$this->secureRandom);
+		$this->encryptAll->encryptAll($input, $output);
```
sharidas: We call encryptAll here. Reusing old code.

sharidas: Since we use encrypt all. Let me also see if we can reuse decrypt all.

sharidas: My mistake, its better to stick on with the decrypt implemented here. The final output of a decryptall is again encrypted. Which is not the result we want.

PVince81: can you add a comment to clarify this, in case future developers wonder about the same question ?

lib/Command/RecreateMasterKey.php
```diff
@@ -0,0 +1,326 @@
...
+				$output->writeln("The process is abandoned");
+			}
+		} else {
+			$output->writeln("Master key is not enabled.\n");
+		}
+	}
+
+	protected function encryptAllUsers(InputInterface $input, OutputInterface $output) {
+		/*
+		 * We are reusing the encryptAll code but not the decryptAll. The reason being
```
sharidas: Added doc which says why we are not using decryptAll and re-using encryptAll. This should help future devs understand why we used this approach.

sharidas:
Testing done as follows:
- [x] Created user ```admin``` and created files.
- [x] Tried to modify the files.
- [x] Ran the command. Relogin as admin user and verified the files were readable.
- [x] Created another user ```user1``` and did the same procedure on both ```admin``` and ```user1```

sharidas:
Closing this PR. We have #12 to track.

PR #8, opened 2017-07-27
Change required for move operation

sharidas:
This change is required so that move
operation executed by transer-ownership
command.

### DIFF
--- a/lib/Crypto/Encryption.php
+++ b/lib/Crypto/Encryption.php
@@ -164,12 +164,14 @@ class Encryption implements IEncryptionModule {
 	 * @param string $mode php stream open mode
 	 * @param array $header contains the header data read from the file
 	 * @param array $accessList who has access to the file contains the key 'users' and 'public'
+	 * @param string|null $sourceFileOfRename Either false or the name of source file to be renamed.
+	 * 						This is helpful for revision increment during move operation between storage.
 	 *
 	 * @return array $header contain data as key-value pairs which should be
 	 *                       written to the header, in case of a write operation
 	 *                       or if no additional data is needed return a empty array
 	 */
-	public function begin($path, $user, $mode, array $header, array $accessList) {
+	public function begin($path, $user, $mode, array $header, array $accessList, $sourceFileOfRename = null) {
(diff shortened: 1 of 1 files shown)


lib/Crypto/Encryption.php
```diff
@@ -189,7 +189,23 @@ public function begin($path, $user, $mode, array $header, array $accessList) {
 		// always use the version from the original file, also part files
 		// need to have a correct version number if they get moved over to the
 		// final location
-		$this->version = (int)$this->keyManager->getVersion($this->stripPartFileExtension($path), new View());
+		if (\OC::$server->getSession()->exists('renameFile')) {
```
PVince81: while this might have worked for confirming the fix, we need a clean approach to detect renames.

Storing in the session is an absolute no go. Imagine if there are concurrent file operations where one is a rename, the value stored in the session would affect any other parallel unrelated file transfers and would likely destroy the known universe

PVince81: please check this location https://github.com/owncloud/core/blob/master/lib/private/Files/Storage/Wrapper/Encryption.php#L652, it is likely more suitable for the fix

lib/Crypto/Encryption.php
```diff
@@ -164,12 +164,14 @@ public function getDisplayName() {
...
 	 * @param array $accessList who has access to the file contains the key 'users' and 'public'
+	 * @param bool|string Either false or the name of source file to be renamed. This is helpful
+	 * 						for revision increment during move operation between storage.
 	 *
 	 * @return array $header contain data as key-value pairs which should be
 	 *                       written to the header, in case of a write operation
 	 *                       or if no additional data is needed return a empty array
 	 */
-	public function begin($path, $user, $mode, array $header, array $accessList) {
+	public function begin($path, $user, $mode, array $header, array $accessList, $renameFile = false) {
```
sharidas: Brought a new argument to begin. There are respective changes https://github.com/owncloud/core/pull/28107. Made changes to encryptions fopen. Also to the "ocencryption" context.

PVince81:
Please backport to stable10 (in core repo)

PR #10, opened 2017-08-02
Inttest running on encryption

SergioBertolinSG:
Trying to run the tests which used to run in jenkins nightly in this repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,9 @@ script:
   - cd tests
   - sh -c "if [ '$INTEGRATION' != '1' -a '$JSTESTS' != '1' -a '$CODECHECK' != '1' -a '$CODECHECK' != '2' ]; then ../../../lib/composer/bin/phpunit --configuration phpunit.xml; fi"
   - cd ..
+  # Run integration tests
+  - sh -c "if [ '$INTEGRATION' = 'encryption' ]; then cd tests/integration; OC_TEST_ALT_HOME=1 OC_TEST_ENCRYPTION_ENABLED=1 ./run.sh; fi"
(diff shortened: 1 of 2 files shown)


SergioBertolinSG:
Travis says: "The job exceeded the maximum time limit for jobs, and has been terminated."

There is a way to increase the time limit?

phil-davis:
I don't think so - Travis puts a 50 minute limit on each job. We had this trouble with UI tests and split them into suites in separate jobs. But I am surprised that these tests would run that long. I see that the Travis log does not have any places that it stops for a long time, so it seems that it really did execute that long.

This ran on Jenkins in core, didn't it?
I guess the Jenkins machine was faster than the VMs that Travis provides.

SergioBertolinSG:
Taking some data from jenkins, the last correct execution of the suite using encryption took ~ 90 minutes. (Without encryption ~43m).

So I see we have several options: 
  - Run these tests in private travis.com, according to travis docs the limit there is 120m https://docs.travis-ci.com/user/customizing-the-build/#Build-Timeouts .
 - Run these tests in private drone.
 - Run there tests in public jenkins again somehow.
 - Improve the core test suite to be able to split the contexts and run them in separated jobs (not immediate solution and needs to allocate time for doing it).

DeepDiver1975:
> Run these tests in private travis.com, according to travis docs the limit there is 120m https://docs.travis-ci.com/user/customizing-the-build/#Build-Timeouts .

not an option - travis-ci.com is for private repos

DeepDiver1975:
> Run these tests in private drone.

lets walk this path - but we need to wait for @tboerger

PVince81:
moving to "planned" for the next iteration

patrickjahns:
Testing with drone to be done here #30

PVince81:
as I understand this is obsoleted by https://github.com/owncloud/encryption/pull/30

PR #14, opened 2017-09-08
Show encryption only if app is enabled

sharidas:
Show encryption in the settings panel
both personal and admin only if the app
is enabled.


tomneedham:
I thought the encryption app was just a default implementation? And that in fact other apps could implement the apis and present a module? In which case doesn't this need to remain in core?

sharidas:
This is with respect to the UI 
encryption enabled -> 
encryption_enabled

encryption disabled -> 
encryption_disabled

sharidas:
This will screw up if there are other encryption apps. Well, in that case then the PR is meaningless. So for now I am closing the PR.

PR #15, opened 2017-09-13
Fix migration issue caused during upgrade to version 10

sharidas:
This change will help users to fix the migration
issue caused during upgrade to version 10. If
user had opted user specific key in the older versions
then no drop down will appear in the version 10.
Same applies if user had opted for the masterkey
in older version, no drop down will appear in version
10.

### DIFF
--- a/appinfo/Migrations/Version20170913113840.php
+++ b/appinfo/Migrations/Version20170913113840.php
@@ -1,0 +1,48 @@
+<?php
+/**
+ * @author Sujith Haridasan <sharidasan@owncloud.com>
+ *
+ * @copyright Copyright (c) 2017, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Encryption\Migrations;
+
+use OCP\Migration\ISimpleMigration;
+use OCP\Migration\IOutput;
+
+class Version20170913113840 implements ISimpleMigration {
+
+	/**
+	 * @param IOutput $out
+	*/
+	public function run(IOutput $out) {
+		$installedVersion = \OC::$server->getConfig()->getSystemValue('version', '0.0.0');
+
+		if (version_compare('10.0.3', $installedVersion, '>=') === true) {
+			$encryptionEnable = \OC::$server->getAppConfig()->getValue('encryption', 'enabled', 'yes');
+			$coreEncryptionEnable = \OC::$server->getAppConfig()->getValue('core', 'encryption_enabled', 'yes');
+			$userSpecificKey = \OC::$server->getAppConfig()->getValue('encryption', 'userSpecificKey', '');
+			$masterKey = \OC::$server->getAppConfig()->getValue('encryption', 'useMasterKey', '');
+
+			if (($userSpecificKey === '') && ($masterKey === '') &&
+				($encryptionEnable === 'yes') && ($coreEncryptionEnable === 'yes')) {
+				\OC::$server->getConfig()->setAppValue('encryption', 'userSpecificKey', '1');
+			}
+		}
+	}
+}
+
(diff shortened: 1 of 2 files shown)


sharidas:
The test matrix is as follows:
- [x] When encryption is enabled with userspecifickey in version 8.2.11 and then upgraded to 10.0.3. Result is : No drop down list for masterkey or userspecific key selection. Only options for recovery key is visible in UI.
- [x] when encryption is enabled with masterkey in version 8.2.11 and then upgraded to version 10.0.3. Result is: No drop down list for masterkey or userspecific key selection. 
- [x] When encryption is enabled with userspecifickey in version 9.1.6 and then upgraded to 10.0.2. Result is : No drop down list for masterkey or userspecific key selection. Only options for recovery key is visible in UI.
- [x] when encryption is enabled with masterkey in version 9.1.6 and then upgraded to version 10.0.2. Result is: No drop down list for masterkey or userspecific key selection.

appinfo/Migrations/Version20170913113840.php
```diff
@@ -0,0 +1,48 @@
...
+class Version20170913113840 implements ISimpleMigration {
+
+	/**
+	 * @param IOutput $out
+	*/
+	public function run(IOutput $out) {
+		$installedVersion = \OC::$server->getConfig()->getSystemValue('version', '0.0.0');
+
+		if ($installedVersion !== '0.0.0') {
+			$maxVersion = explode('.', $installedVersion)[0];
```
PVince81: there's a more PHP-y way to do this, see https://github.com/owncloud/core/blob/master/lib/private/Repair/Apps.php#L99

appinfo/Migrations/Version20170913113840.php
```diff
@@ -0,0 +1,46 @@
...
+
+class Version20170913113840 implements ISimpleMigration {
+
+	/**
+	 * @param IOutput $out
+	*/
+	public function run(IOutput $out) {
+		$installedVersion = \OC::$server->getConfig()->getSystemValue('version', '0.0.0');
+
+		if (version_compare('10.0.0', $installedVersion) === 1) {
```
sharidas: This looks better.

PVince81:
@sharidas when updating from 10.0.2 the logic should also operate, since 10.0.2 currently still shows the dropdown. When I upgrade from 10.0.2 with working encryption but dropdown shown to 10.0.3, then dropdown should disappear. It should be possible by running the same logic.

appinfo/Migrations/Version20170913113840.php
```diff
@@ -0,0 +1,46 @@
...
+
+class Version20170913113840 implements ISimpleMigration {
+
+	/**
+	 * @param IOutput $out
+	*/
+	public function run(IOutput $out) {
+		$installedVersion = \OC::$server->getConfig()->getSystemValue('version', '0.0.0');
+
+		if (version_compare('10.0.0', $installedVersion) !== 0) {
```
sharidas: Verified with the change by upgrading from 10.0.2 to 10.0.3. In both the user specific key or master key when enabled the later version doesn't show the dropdown list now.

PVince81: what you wrote here is basically `$version !== '10.0.0'` not `$version <= 10.0.3`

The reason it works is probably because the check in your "if" block would likely work correctly even without a version check. I'd rather have either the correct logic with version check, or no version check at all.

Furthermore if someone upgrades from 10.0.0 the logic below will not run at all.
Please recheck the documentation of `version_compare`

appinfo/Migrations/Version20170913113840.php
```diff
@@ -0,0 +1,48 @@
...
+
+class Version20170913113840 implements ISimpleMigration {
+
+	/**
+	 * @param IOutput $out
+	*/
+	public function run(IOutput $out) {
+		$installedVersion = \OC::$server->getConfig()->getSystemValue('version', '0.0.0');
+
+		if (version_compare('10.0.3', $installedVersion, '>=') === true) {
```
sharidas: Now this is the only modification I made. This is equivalent to `10.0.3 >= $installeddVersion`.

Since we are heading towards 10.0.3, this should be good enough. The UI changed from 10.0.2.
So till version < 10.0.2, the upgrade should work as it did. Or lets say if `userSpecificKey` is added to appconfig for versions less than 10.0.2, then, nothing would prohibit the code flow. 

Now once the upgrade reaches 10.0.2 which is still less than 10.0.3, goes inside the if condition. And then tries to check if the condition is met with the inner `if` condition ( the one with lots of `&&`). And adds the `userSpecificKey` if the condition is true and UI will not have dropdown list.

Lets say if version >= 10.0.2, then the inner if condition may/may not be hit. Because if the encryption is enabled then either of the types would be selected. If not then again falls back to `userSpecificKey`. This logic applies till, including version 10.0.3.

Though technically the version check is redundant imho. Because unless explicitly selected for masterkey or enabled masterkey via command line => its userSpecificKey. Which sticks with the inner if condition.

PVince81:
@sharidas please backport to stable10 for 10.0.4

sharidas:
Backport : https://github.com/owncloud/core/pull/29049

PR #16, opened 2017-10-04
Cleanup alternate storage location keys of user

sharidas:
This is to make sure that there are no user
specific folders if the keys are stored in
alternate location apart from the default
location, when a user is deleted. Hence the
keys inside user folder in the alternate
location would be deleted.

### DIFF
--- a/lib/Hooks/UserHooks.php
+++ b/lib/Hooks/UserHooks.php
@@ -198,8 +198,16 @@ class UserHooks implements IHook {
 	public function postDeleteUser($params) {
 
 		if (App::isEnabled('encryption')) {
-			$this->keyManager->deletePublicKey($params['uid']);
+			/**
+			 * Adding a safe condition to make sure the uid is not
+			 * empty or null.
+			 */
+			if (!is_null($params['uid']) && ($params['uid'] !== '')) {
+				$this->keyManager->deletePublicKey($params['uid']);
+				\OC::$server->getEncryptionKeyStorage()->deleteAltUserStorageKeys($params['uid']);
+			}
 		}
+
 	}
 
 	/**


lib/Hooks/UserHooks.php
```diff
@@ -200,6 +200,8 @@ public function postDeleteUser($params) {
 		if (App::isEnabled('encryption')) {
 			$this->keyManager->deletePublicKey($params['uid']);
 		}
+
+		\OC::$server->getEncryptionKeyStorage()->deleteAltUserStorageKeys($params['uid']);
```
jvillafanez: Could the key storage be injected?

jvillafanez:
Specify what you've tested in the comments in case there are some scenarios that you didn't consider.

Some questions I have:
* What happens in the keys aren't in the alternate location? Code shouldn't explode and just ignore
* What happens if the user uid is null, empty or other weird scenario?

lib/Hooks/UserHooks.php
jvillafanez: Should this piece be executed only if the encryption app is enable? (just asking)

sharidas:
So the behaviour without this change:
- When keys are stored in default location, the deletion of user will clean up the keys associated with the users.
- When keys are changed to some different location ( obviously relative to data folder ), the deletion of user will not clean up the keys associated with the users.
After applying this change:
- The keys are deleted even if the storage location of keys are not default for the users, when the user is deleted.
- If keys are in the default location, code wont explode.

Do we really need to check if the uid exists, if its null/empty? Because all we are trying to achieve here is a simple remove operation if : the storage location is not default one.

jvillafanez:
> Do we really need to check if the uid exists, if its null/empty?

According to what I see in https://github.com/owncloud/core/blob/17691e444641a84103387ef118beee9b0df4f112/lib/private/Encryption/Keys/Storage.php#L197 if somehow the user uid is empty, it could imply that all the user keys are deleted. This is really bad.
Is it possible that such scenario happen? I don't really know, but I'd rather stay on the safe side and make sure it won't ever happen. If you assure that the uid will always contain something and such scenario won't happen, I'm fine if you add a comment there.

sharidas:
I have verified both using UI and command to delete user, uid wasn't found null/empty or so. You are right, if the uid becomes empty then it would delete the entire folder. So I have added a check before calling the method to make sure its not null or empty.

lib/Hooks/UserHooks.php
```diff
@@ -200,6 +200,10 @@ public function postDeleteUser($params) {
 		if (App::isEnabled('encryption')) {
 			$this->keyManager->deletePublicKey($params['uid']);
 		}
+
+		if (!is_null($params['uid']) && ($params['uid'] !== '')) {
```
jvillafanez: I think it's better to move this inside the `App::isEnabled('encryption')` check, just for consistency. Maybe we can remove the `App::isEnabled('encryption')` check and remove the keys in any case. It's weird that some keys are removed if the encryption is disabled but others not.

If this is the right behaviour, add a comment explaining.

sharidas: Moved the piece of code inside `App::isEnabled('encryption')` check. Since its a hook, imho it would not be a good idea to remove the check `App::isEnabled('encryption')`.

PR #21, opened 2017-12-01
Add re-login information after successful action

sharidas:
After the command is executed successfully by
recreating masterkey, the user should be informed
to relogin.

### DIFF
--- a/lib/Command/RecreateMasterKey.php
+++ b/lib/Command/RecreateMasterKey.php
@@ -226,6 +226,7 @@ class RecreateMasterKey extends Command {
 				$this->keyManager->validateMasterKey();
 				$this->encryptAllUsers($input, $output);
 				$output->writeln("\nEncryption completed successfully\n");
+				$output->writeln("\n<info>Note: All users are required to relogin.</info>\n");
 			} else {
 				$output->writeln("The process is abandoned");
 			}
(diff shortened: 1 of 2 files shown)


lib/Command/RecreateMasterKey.php
```diff
@@ -231,6 +231,7 @@ protected function execute(InputInterface $input, OutputInterface $output) {
 				$this->keyManager->validateMasterKey();
 				$this->encryptAllUsers($input, $output);
 				$output->writeln("\nEncryption completed successfully\n");
+				$output->writeln("\nNOTE: Kindly log-out and log-in to access the files encrypted with new master key.");
```
PVince81: "All users will need to log out and log in again".

Not only the admin...

tests/Command/RecreateMasterKeyTest.php
```diff
@@ -197,6 +197,7 @@ function($path) {
 				->willReturn(true);
 
 			global $outputText;
+			global $reloginText;
```
PVince81: don't use globals but `use (&$reloginText)` instead in the closure

sharidas: Done

lib/Command/RecreateMasterKey.php
sharidas: Done

sharidas:
The imho CI failure doesn't look related to this change...

PVince81:
@sharidas please backport

sharidas:
Backport to stable10 : https://github.com/owncloud/core/pull/34596

PR #22, opened 2017-12-04
skip shared files when adding recovery key

butonic:
Skip shared files when adding recovery key.

### DIFF
--- a/lib/Recovery.php
+++ b/lib/Recovery.php
@@ -33,6 +33,7 @@ use OCP\PreConditionNotMetException;
 use OCP\Security\ISecureRandom;
 use OC\Files\View;
 use OCP\Encryption\IFile;
+use OCP\Files\FileInfo;
 
 class Recovery {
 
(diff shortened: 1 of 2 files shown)


lib/Recovery.php
```diff
@@ -221,6 +221,11 @@ private function addRecoveryKeys($path) {
 		$dirContent = $this->view->getDirectoryContent($path);
 		foreach ($dirContent as $item) {
 			$filePath = $item->getPath();
+			// only add key for files owned by the user
+			if($item->getStorage()->instanceOfStorage('OC\Files\Storage\Shared')) {
```
PVince81: `ISharedStorage` to also exclude incoming federated shares.

See and also please review: https://github.com/owncloud/core/pull/29724

PVince81 (changes requested):

lib/Recovery.php
tomneedham: And we can use ::class`

butonic: hedcoding because we dont want inter app dependencies. might give problems during class loading in case files_sharing is disabled

PVince81:
@sharidas please take over

sharidas:
I have updated the PR by adding a unit test. I have verified this PR with normal share (i.e, sharing a file/folder within the same oC instance) and federated share. The verification was done to see the code flow. And in both cases when the recovery is enabled for user, it was hitting code https://github.com/owncloud/encryption/pull/22/files#diff-af3b60507c206ac40d8ec514b7cb4f63R344.

Ran the unit test locally and it did not caused any issue.

PVince81 (approved):
:+1:

phil-davis:
Backport to core `stable10` is PR https://github.com/owncloud/core/pull/34506

PR #30, opened 2018-03-29
Init unit testing in drone

patrickjahns:
# Motivation

Boostrap testing of encryption app in drone

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -6,10 +6,12 @@ branches: [master, release*]
 
 pipeline:
   install-server:
-    image: owncloudci/core
+    image: patrickjahns/occore
     pull: true
     exclude: "apps/encryption"
     version: ${OC_VERSION}
+    db_type: ${DB_TYPE}
+    db_host: ${DB_HOST}
 
   code-check:
     image: owncloudci/php:${PHP_VERSION}
@@ -19,12 +21,20 @@ pipeline:
       #- cd /var/www/owncloud
       #- php ./occ app:check-code encryption -c private -c strong-comparison
       #- php ./occ app:check-code encryption -c deprecation
+    when:
+      matrix:
+        TEST_SUITE: phpunit
+
 
   phpunit:
     image: owncloudci/php:${PHP_VERSION}
     pull: true
     commands:
-      - ../../lib/composer/bin/phpunit --configuration tests/phpunit.xml
+      - ../../lib/composer/bin/phpunit --configuration tests/phpunit.xml --log-junit tests/output/junit/report2.xml
+    when:
+      matrix:
+        TEST_SUITE: phpunit
+
 
   configure-server:
     image: owncloudci/php:${PHP_VERSION}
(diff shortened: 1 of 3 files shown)


phil-davis:
That looks good. Just some "remaining quota" tests fail like:
```
webdav-related-new-endpoint.feature:134
    Then the single response should contain a property "{DAV:}quota-available-bytes" with value "10485358" # FeatureContext::theSingleResponseShouldContainAPropertyWithValue()
Property "{DAV:}quota-available-bytes" found with value "10485429", expected "10485358" (Exception)
```
this, and the 3 scenarios after it, are too brittle. The remaining quota depends on the exact size of files in the skeleton and what the encryption happens to do with those file's sizes.

patrickjahns:
It's moving forward - however when enabling user-key encryption more tests fail (as expected)

I've not yet limited the flags or let the `run.sh` script set certain things - just enforced master/userkey encryption by hand

I've tried to run also the UI acceptance tests - but they had some failures - let me enable them for a round to give some feedback of what might happen there

patrickjahns:
@phil-davis 
Is there any major difference in master vs stable10 in regards to ui tests - stable10 fails after around 1:15 - but master runs into a timeout at 2hours

phil-davis:
The infrastructure is the same. The content difference is just that ``stable10`` has the test suites ``webUIManageUsersGroups`` and ``webUIManageQuota`` but in ``master`` those have moved out into the ``user_management`` app.

So actually I expect ``master`` to run quicker.

patrickjahns:
All UI tests are ✅ - that's great. I guess some of the ui tests don't make sense to run here. Should we just limit the tests to the ones that actually deal with files ?

API Test errors need investigation - and I also think some of them might not be useful on encryption (like quota)

phil-davis:
Probably you can just sort out setting the appropriate external vars like:
```
OC_TEST_ENCRYPTION_ENABLED
OC_TEST_ENCRYPTION_MASTER_KEY_ENABLED
OC_TEST_ENCRYPTION_USER_KEYS_ENABLED
```
that filter out some test scenarios by selecting or not by tags already in the scenarios - this code in ``run.sh`` controls it:
```
# Enable encryption if requested
if test "$OC_TEST_ENCRYPTION_ENABLED" = "1"; then
	env_encryption_enable
	BEHAT_FILTER_TAGS="~@no_encryption&&~@no_default_encryption"
elif test "$OC_TEST_ENCRYPTION_MASTER_KEY_ENABLED" = "1"; then
	env_encryption_enable_master_key
	BEHAT_FILTER_TAGS="~@no_encryption&&~@no_masterkey_encryption"
elif test "$OC_TEST_ENCRYPTION_USER_KEYS_ENABLED" = "1"; then
	env_encryption_enable_user_keys
	BEHAT_FILTER_TAGS="~@no_encryption&&~@no_userkeys_encryption"
fi
```
And that will stop a lot of these scenarios being run with encryption.

patrickjahns:
I've had a look at it before - but would also like to go through the scenarios once, and see which have been excluded on encryption ( and why ). This is not urgent, but as much as I can see, cleanup of the acceptance suites is also not only of the (duplicated) behat steps. 

With these flags, there is also the question, why we expose 3 settings ? Whats the difference between "default_encryption" - and the other two modes. 
How will they affect the scenarios ?

phil-davis:
I had a quick look and do not see any scenarios tagged ``no_masterkey_encryption`` or ``no_userkeys_encryption`` so it looks like those have not been used.

patrickjahns:
@phil-davis @DeepDiver1975 @PVince81 @individual-it 
Changed the approach to first getting unit tests merged into the repository 

Will create two seperate PRs for API / UI tests

phil-davis:
```
There was 1 warning:

1) OCA\Encryption\Tests\Crypto\EncryptAllTest::testEncryptUsersFiles
Class "Symfony\Component\Console\Helper\ProgressBar" is declared "final" and cannot be mocked.
```
The same warning is in Travis and drone logs, so not introduced here. I am just noting it because in ``core`` we recently made it so that warnings would fail the job.

PR #34, opened 2018-05-27
Bootstrap UI testing on drone

patrickjahns:
Based on top of #30 - edit: which has been merged and this is now rebased on master

Initializes UI testing within drone

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -239,7 +239,7 @@ matrix:
       DB_TYPE: sqlite
 
     # Acceptance Tests
-    ## UI
+    ## UI core stable10 with masterkey encryption
     - PHP_VERSION: 5.6
       OC_VERSION: daily-stable10-qa
       PREPARE_ACCEPTANCE: true
@@ -336,6 +336,7 @@ matrix:
       DB_HOST: mysql
       PART: 8
 
+    ## UI core stable10 with user-keys encryption
     - PHP_VERSION: 5.6
       OC_VERSION: daily-stable10-qa
       PREPARE_ACCEPTANCE: true
@@ -432,6 +433,7 @@ matrix:
       DB_HOST: mysql
       PART: 8
 
+    ## UI core master with masterkey encryption
     - PHP_VERSION: 7.1
       OC_VERSION: daily-master-qa
       PREPARE_ACCEPTANCE: true
@@ -528,6 +530,7 @@ matrix:
       DB_HOST: mysql
       PART: 8
 
+    ## UI core master with user-keys encryption
     - PHP_VERSION: 7.1
       OC_VERSION: daily-master-qa
       PREPARE_ACCEPTANCE: true
--- a/.drone.yml
+++ b/.drone.yml
@@ -13,6 +13,41 @@ pipeline:
     db_type: ${DB_TYPE}
     db_host: ${DB_HOST}
 
+  install-fed-server:
+    image: owncloudci/core
+    pull: true
+    version: ${FEDERATION_OC_VERSION}
+    core_path: /var/www/owncloud/fed-server
+    when:
+      matrix:
+        USE_FEDERATED_SERVER: true
+
+  configure-federation-server:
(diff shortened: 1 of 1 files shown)


phil-davis:
It fails a webUI test that tries to do a password reset for a user.
In the encryption case, is there a different sequence of actions for the user resetting their password?

sharidas:
- When user-keys is configured for encryption in oC instance and password recovery option is enabled: then user would see a page like this in Users page logged in as `admin`: 
userkeys_password_recovery
- Without recovery password the UI would look default users page: 
userkeys_without_recoverypassword
- with masterkey enabled users page would look as default page: 
masterkey_userspage

Hope this helps?

patrickjahns:
@sharidas 
It is rather about when a user self-resets his/her password - as he has forgotten his password

patrickjahns:
@phil-davis 

quickly tested it - the login screen has a confirm page - see:

image

.drone.yml
```diff
@@ -61,6 +61,48 @@ pipeline:
...
+      - REMOTE_FED_SRV_HOST_NAME=federated
+      - REMOTE_FED_SRV_HOST_PORT=80
+      - SKELETON_DIR=/var/www/owncloud/tests/acceptance/webUISkeleton
+      - SELENIUM_PORT=4444
+      - PLATFORM=Linux
+      - MAILHOG_HOST=email
+    commands:
+      - cd /var/www/owncloud
+      - chmod 777 /var/www/owncloud/tests/acceptance/filesForUpload -R
+      - bash tests/travis/start_ui_tests.sh --remote --all-suites
```
phil-davis: Now that https://github.com/owncloud/core/pull/31808 is in core, it should be possible to skip core tests that are different or not relevant for encryption by adding:
```
--tags '~@skipOnEncryption'
```
in a similar way to the ``.drone.yml`` in the ``user_ldap`` app.

patrickjahns: will push that later tonight / tomorrow - once the tags are in stable10 / master

phil-davis:
@patrickjahns encryption is still bundled in core stable10. It "magically appears" in the core apps folder  when I ``git checkout stable10`` from core. Some discussion in 2017 https://github.com/owncloud/core/pull/29007

So testing the separate ``encryption`` app against ``daily-stable10-qa`` is not really a valid thing.

That will reduce the number of drone jobs that can fail.

phil-davis:
@patrickjahns my point was that the encryption app here in a separate repo is the encryption app that has been unbundled from core master. At the moment, it can't be added/installed to a core ``stable10`` system, because that system already has a folder ``apps/encryption`` with the encryption app code in it.

But maybe I misunderstand. Maybe when a test suite runs here against a core ``stable10`` system, it is using the encryption code already in that core ``stable10`` system (and not overwriting with this encryption app code)???

patrickjahns:
@phil-davis 
It runs this app-code against stable10

https://github.com/owncloud/encryption/blob/66ebfc76212c11aea5800c2484a3efa43a16abe8/.drone.yml#L11

sharidas:
> So testing the separate encryption app against daily-stable10-qa is not really a valid thing.

@patrickjahns @phil-davis Out of curiosity, I would like to know what if a change happens in the core ( master branch ), and effectively changed in encryption app ( master branch ). And this encryption app would run CI agains stable10 branch of core ( daily-stable10-qa ). 

Now lets think of the situation, the above mentioned change in core ( master branch ) is merged. Then CI for stable10 might fail, right? Because stable10 will not have the change merged to master. This is thought though.

phil-davis:
I added a small commit to put a comment before each block of 8-part matrix entries for the UI tests. That makes it easier to see what is going on with the now-long list of matrix entries.

That also triggers CI, which is good. Because the core code that handles auto-splitting the available suites into parts was fixed/enhanced on Friday. That all worked nicely in ``files_primary_s3`` CI runs earlier today. So it will be good to see the results here.

PR #36, opened 2018-06-01
Enable user-keys if no encryption mode is enabled during login

sharidas:
Enable user-keys if no encryption mode is enabled
during user login.


sharidas:
Created this PR based on the request from https://github.com/owncloud/core/pull/31597.

PR #37, opened 2018-06-18
initialize ui testing

phil-davis:


phil-davis:
Hmmm - I missed pushing the little change I was making. It is day-time in Europe now, so no point running this to "get ahead" with a CI run.

PR #40, opened 2018-07-26
Provision to access recovery and oc password from env

sharidas:
It should be able to read recovery password and oc
password from the env. This change helps to do the
same.

### DIFF
--- a/lib/Crypto/DecryptAll.php
+++ b/lib/Crypto/DecryptAll.php
@@ -23,6 +23,7 @@
 namespace OCA\Encryption\Crypto;
 
 
+use OC\Helper\EnvironmentHelper;
 use OC\User\LoginException;
 use OCA\Encryption\KeyManager;
 use OCA\Encryption\Session;
@@ -42,6 +43,9 @@ class DecryptAll {
 	/** @var QuestionHelper  */
 	protected $questionHelper;
 
+	/** @var EnvironmentHelper  */
+	protected $environmentHelper;
+
 	/** @var  Crypt */
 	protected $crypt;
 
@@ -55,11 +59,15 @@ class DecryptAll {
 	protected $userManager;
 
 	/**
+	 * DecryptAll constructor.
+	 *
 	 * @param Util $util
 	 * @param KeyManager $keyManager
 	 * @param Crypt $crypt
 	 * @param Session $session
+	 * @param IUserManager $userManager
 	 * @param QuestionHelper $questionHelper
+	 * @param EnvironmentHelper $environmentHelper
 	 */
 	public function __construct(
 		Util $util,
@@ -67,7 +75,8 @@ class DecryptAll {
 		Crypt $crypt,
 		Session $session,
 		IUserManager $userManager,
-		QuestionHelper $questionHelper
+		QuestionHelper $questionHelper,
+		EnvironmentHelper $environmentHelper
 	) {
 		$this->util = $util;
 		$this->keyManager = $keyManager;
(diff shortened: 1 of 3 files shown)


lib/Crypto/DecryptAll.php
```diff
@@ -90,55 +90,77 @@ public function __construct(
...
+			// Must either be using recovery keys, or just decrypting one user and know their password
+
+			// Check a method has been passed
+			if(!$input->hasOption('method') && !in_array($input->getOption('method'),['recovery', 'password'])) {
+				$output->writeln('A method must be supplied when decrypting from user-key state');
+				return false;
+			}
+			if (empty($user)) {
+				// all users, so only recovery is possible
+				if($input->getOption('method')==='recovery' && !empty(getenv('OC_RECOVERY_PASSWORD'))) {
```
PVince81: please avoid using `empty` because in PHP `empty("0")` is true...

PVince81: also please add more spacing

PVince81: and you missed the backslash before `getenv` here

lib/Crypto/DecryptAll.php
```diff
@@ -90,55 +90,77 @@ public function __construct(
...
-				try {
-					if ($this->keyManager->checkRecoveryPassword($password) === false) {
+					if ($this->userManager->checkPassword($user, $password) === false) {
 						$throwLoginException = true;
 					}
-				} catch (\Exception $e) {
-					$throwLoginException = true;
+				} elseif ($input->getOption('method') === null) {
+					//Grab user specific password
+					$question = new Question('Please enter the user: ' . $user . ' login password: ');
```
PVince81: "Please enter the login password from user $user: "

lib/Crypto/DecryptAll.php
```diff
@@ -149,12 +171,12 @@ public function prepare(InputInterface $input, OutputInterface $output, $user) {
 		$privateKey = $this->getPrivateKey($user, $password);
 		if ($privateKey !== false) {
 			$this->updateSession($user, $privateKey);
+			$output->writeLn('Setup encryption module ready for decryption');
```
PVince81: this looks more like debugging information, not sure we should show this to an admin...

tests/Crypto/DecryptAllTest.php
```diff
@@ -182,17 +186,184 @@ public function testPrepareLoginException($loginType) {
...
+			->willReturn(false);
+
+		$input->expects($this->once())
+			->method('hasOption')
+			->willReturn(true);
+		$input->expects($this->exactly(2))
+			->method('getOption')
+			->with('method')
+			->willReturnOnConsecutiveCalls(null, 'password');
+		\putenv("OC_PASSWORD=foo");
```
PVince81: how about extending core's "EnvironmentHelper" and adding a method for reasing env variables ?

this means you'd be able to mock it in the tests

lib/Crypto/DecryptAll.php
sharidas: Addressed the spacing, backslash before `getenv` and removed usage of `empty`.

lib/Crypto/DecryptAll.php
sharidas: Done.

lib/Crypto/DecryptAll.php
sharidas: Removed as of now.

tests/Crypto/DecryptAllTest.php
sharidas: Added a method in EnvironmentHelper. And in the tests its being mocked.

PVince81:
please clearly state in the PR first post that this PR requires the other core one

lib/Crypto/DecryptAll.php
```diff
@@ -90,55 +100,98 @@ public function __construct(
...
 	public function prepare(InputInterface $input, OutputInterface $output, $user) {
-
-		$question = new Question('Please enter the recovery key password: ');
-
+		// We need to setup the module with everything it needs in order to decrypt user data
+		// This could be using master key or user key encryption
+		// and could also be for a single user or multiple users
+
+		//Try to stop decrypting process if password method is used for decrypting all users.
+		if (($input->getArgument('user') === '') && ($input->getOption('method') === 'password')) {
```
PVince81: why check the user argument in here when it's already passed as `$user` ?

sharidas:
This PR should be reviewed along with core PR: https://github.com/owncloud/core/pull/32172

lib/Crypto/DecryptAll.php
sharidas: Thought of adding an extra precaution.

sharidas: I have updated the comments in the latest version of the PR. Let me know if the description looks ok?

lib/Crypto/DecryptAll.php
```diff
@@ -90,55 +100,104 @@ public function __construct(
...
+		// and could also be for a single user or multiple users
+
+		/**
+		 * Try to stop decrypting process if password method is used for decrypting all users.
+		 * If the decrypt-all command doesn't provide a user, which means decryption of oC
+		 * filesystem. And if the method provided is password then it should stop decryption
+		 * process, because same password can not be used for every user. Even though the
+		 * input user is empty the $user can have value of each user(s) in oC.
+		 */
+		if (($input->getArgument('user') === '') && ($input->getOption('method') === 'password')) {
```
PVince81: please use `empty($user)` here like it's done further below.

then also update the PHPDoc of this method to have the proper types for the "$user" argument as it's not clear whether it's a string or not. also it's not clear that it can be null and that null means "all"

sharidas: There is a small problem I see here to use `empty($user)` to satisfy the condition what I am trying to achieve. The `$user` is not empty considering the code path triggered from `encryption:decrypt-all` command. 

Let say user-keys encryption is enabled and admin wants to run decrypt-all on all users. And the command triggered was:
`./occ encryption:decrypt-all -m password -c yes` 
This would reach the line
- https://github.com/owncloud/core/blob/master/lib/private/Encryption/DecryptAll.php#L176
- which would eventually call `prepare()` and thus prepare will have uid of the user passed. So that way `$user` the uid cannot be empty :(

The place where `$user` becomes `""` is while decrypting masterkey encrypted files.

PVince81: so does it mean that the check for `empty($user)` on line 137 is never called either ?

PVince81: ok, please update the PHPDoc of the method to say so: when is $user an empty string and what does it mean.
this place is already confusing enough

sharidas: I have updated the PHPDoc of the method.

lib/Crypto/DecryptAll.php
```diff
@@ -55,90 +59,149 @@ class DecryptAll {
...
-
-		$question = new Question('Please enter the recovery key password: ');
-
+		// We need to setup the module with everything it needs in order to decrypt user data
+		// This could be using master key or user key encryption
+		// and could also be for a single user or multiple users
+
+		/**
+		 * Try to stop decrypting process if password method is used for decrypting all users.
+		 * If the decrypt-all command doesn't provide a user, which means decryption of oC
```
PVince81: this is too long, no one will read this. PHPdoc needs to say that $user can be null and tell in what situation

lib/Crypto/DecryptAll.php
```diff
@@ -55,90 +59,149 @@ class DecryptAll {
...
+	 *   the uid of user passed, if the oC filesystem is encrypted using user-keys
+	 *   the empty string '', if the oC filesystem is encrypted using master-key
+	 *
 	 * - Throws LoginException when user login fails either recovery password fails
 	 *   or if the user password fails
 	 *
 	 * @param InputInterface $input
 	 * @param OutputInterface $output
-	 * @param $user
+	 * @param string|null $user The uid of the user ( a non empty string ) for user-keys, or empty or null string for masterkey decryption
```
sharidas: Updated the description for `$user` at `@param`.

PVince81:
@ownclouders rebase

PR #41, opened 2018-08-13
run api acceptance tests from core

individual-it:
run all API tests from core with encryption app enabled

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -154,7 +154,7 @@ pipeline:
       - SELENIUM_PORT=4444
       - PLATFORM=Linux
       - MAILHOG_HOST=email
-      - DIVIDE_INTO_NUM_PARTS=8
+      - DIVIDE_INTO_NUM_PARTS=12
       - RUN_PART=${PART}
     commands:
       - cd /var/www/owncloud/tests/acceptance/
@@ -164,6 +164,23 @@ pipeline:
     when:
       matrix:
         TEST_SOURCE: core-webUI
+
+  core-api-acceptance-tests:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    environment:
+      - TEST_SERVER_URL=http://server
+      - MAILHOG_HOST=email
+      - DIVIDE_INTO_NUM_PARTS=12
+      - RUN_PART=${PART}
+    commands:
+      - cd /var/www/owncloud/tests/acceptance/
+      - touch /var/www/owncloud/saved-settings.sh
+      - . /var/www/owncloud/saved-settings.sh
+      - su-exec www-data ./run.sh --remote --type api --tags '~@skipOnEncryption&&~@skipOnEncryptionType:${ENCRYPTION_TYPE}&&~@skip'
+    when:
+      matrix:
+        TEST_SOURCE: core-api
 
   slack:
     image: plugins/slack:1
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -625,4 +643,392 @@ matrix:
...
+      DB_HOST: mysql
+      PART: 8
+
+    ## API core master with masterkey encryption
+    - PHP_VERSION: 7.1
+      OC_VERSION: daily-master-qa
+      PREPARE_ACCEPTANCE: true
+      USE_SERVER: true
+      USE_FEDERATED_SERVER: true
+      FEDERATION_OC_VERSION: daily-stable10-qa
```
phil-davis: still says stabe10 here - 8 parts

.drone.yml
```diff
@@ -625,4 +643,392 @@ matrix:
...
+      DB_HOST: mysql
+      PART: 8
+
+    ## API core master with user-keys encryption
+    - PHP_VERSION: 7.1
+      OC_VERSION: daily-master-qa
+      PREPARE_ACCEPTANCE: true
+      USE_SERVER: true
+      USE_FEDERATED_SERVER: true
+      FEDERATION_OC_VERSION: daily-stable10-qa
```
phil-davis: still says stabe10 here - 8 parts

.drone.yml
individual-it: that is the federation server. currently the fed. servers in all tests is running stable10. Encryption is not enabled on the fed. server and we want to make sure that any change can still talk to an existing server in the wild. stable10 is the closest we can get to a possible version in the wild

phil-davis: Good, glad you are thinking :)

phil-davis:
https://drone.owncloud.com/owncloud/encryption/147/860
```
--- Failed scenarios:

    /var/www/owncloud/tests/acceptance/features/apiTrashbin/trashbin-new-endpoint.feature:197
    /var/www/owncloud/tests/acceptance/features/apiTrashbin/trashbin-old-endpoint.feature:186
    /var/www/owncloud/tests/acceptance/features/apiTrashbin/trashbin-old-endpoint.feature:197
    /var/www/owncloud/tests/acceptance/features/apiTrashbin/trashbin-old-endpoint.feature:215
```

patrickjahns:
> CI just needs to pass - there are some trashbin user_keys encryption combinations that are failing and already have some skip tags. They probably will need to be skipped here also?

@sharidas - could you confirm what is allowed to be skipped

individual-it:
there are more issues with failing tests. one is this https://github.com/owncloud/core/issues/32322
but for me it looks like to behaviour with encryption is correct

sharidas:
I did a manual test in stable10 with user-key encryption on, the deletion of file from external storage ( for example sftp ) moves the file to trashbin. And restore brings it back to the external storage. 
I believe the encryption is working as expected.

individual-it:
an other problem is the quota 
`'Exception' with message 'Property "{DAV:}quota-available-bytes" found with value "10485429", expected "10485358"'`
`Exception' with message 'Property "{DAV:}quota-available-bytes" found with value "600", expected "529"'`
`'Exception' with message 'Property "{DAV:}quota-available-bytes" found with value "693", expected "622"'`

with encryption there are always 71bytes more quota available that without, but cannot reproduce it locally

individual-it:
`.gitignore` file is exactly 71 bytes big

phil-davis:
code/script changes still look good. Can be merged if CI passes.

phil-davis:
We want to know that file systems work (local, object store, WND, ...) when plugged in behind the middleware ownCloud code. So having some sort of abstraction used in testing that might behave a bit differently from "the real thing in production" is risky.

individual-it:
I think the problem is not the "real" filesystem but the test that expects an exact match, should make the tests more flexible one day

phil-davis:
> should make the tests more flexible one day

make them always return true, that will save as from annoying test fails ;)

individual-it:
drone job hang for an unknown reason, retriggered

individual-it:
so annoying, drone failed this time because of problems in the QA build
image
restarted

individual-it:
now codecov failed 
image
restarting

PR #48, opened 2018-08-24
Update README.md

mmattel:
Initial readme.md text copied from info.xml

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,3 +4,12 @@
  [![Build Status](https://travis-ci.org/owncloud/encryption.svg?branch=master)](https://travis-ci.org/owncloud/encryption)
  [![Scrutinizer Code Quality](https://scrutinizer-ci.com/g/owncloud/encryption/badges/quality-score.png?b=master)](https://scrutinizer-ci.com/g/owncloud/encryption/?branch=master)
  [![Code Coverage](https://scrutinizer-ci.com/g/owncloud/encryption/badges/coverage.png?b=master)](https://scrutinizer-ci.com/g/owncloud/encryption/?branch=master)
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,3 +4,12 @@
  [![Build Status](https://travis-ci.org/owncloud/encryption.svg?branch=master)](https://travis-ci.org/owncloud/encryption)
  [![Scrutinizer Code Quality](https://scrutinizer-ci.com/g/owncloud/encryption/badges/quality-score.png?b=master)](https://scrutinizer-ci.com/g/owncloud/encryption/?branch=master)
  [![Code Coverage](https://scrutinizer-ci.com/g/owncloud/encryption/badges/coverage.png?b=master)](https://scrutinizer-ci.com/g/owncloud/encryption/?branch=master)
+
+In order to use this encryption module you need to enable server-side
+encryption in the admin settings. Once enabled this module will encrypt
+all your files transparently. The encryption is based on AES 256 keys.
+The module won't touch existing files, only new files will be encrypted
+after server-side encryption was enabled. It is also not possible to
+disable the encryption again and switch back to a unencrypted system.
```
PVince81: it is possible if you use a recovery key or master key and `occ decrypt-all`, but the setup is currently complicated as you need every user to opt-in

mmattel: As said, I have copied the text from info.xml
Please make a (as exact as possible) text proposal, and I will change it in readme.md and info.xml

In any case, if it would take too long to change the text here, I propose to merge this PR and create a second one with changes for both files having them in sync.

PVince81: ok agreed, let's merge

Issue #54, opened 2018-09-12
Acceptance test failures related to add user by email address


phil-davis:
Some acceptance tests are failing. See example in PR #53 and its linked drone results https://drone.owncloud.com/owncloud/encryption/205

e.g. with user-keys, mysql, daily-stable10-qa PHP 5.6
```
 Scenario Outline: user sets his own password after being created with an Email address only # /var/www/owncloud/tests/acceptance/features/webUIManageUsersGroups/addUsers.feature:61
      | guiusr1  | simple user-name      |
        Failed asserting that '
        --_=_swift_v4_1536720919_49c4e00e6c370c3071e621950db80cd9_=_
        Content-Type: text/plain; charset=utf-8
        Content-Transfer-Encoding: quoted-printable
        
        Hey there,
        
        just letting you know that you now have an ownCloud account.
        
        Your username: guiusr1
        Access it: http://server/settings/users/setpassword/form/761553510911067257740/guiusr1
        
        Cheers!
        	--
        ownCloud - A safe home for all your data
        https://owncloud.org
        
        --_=_swift_v4_1536720919_49c4e00e6c370c3071e621950db80cd9_=_
        Content-Type: text/html; charset=utf-8
        Content-Transfer-Encoding: quoted-printable
        
        <table cellspacing="0" cellpadding="0" border="0" width="100%">
        	<tr><td>
        			<table cellspacing="0" cellpadding="0" border="0" width="600px">
        				<tr>
        					<td bgcolor="#1d2d44" width="20px">&nbsp;</td>
        					<td bgcolor="#1d2d44">
        						<img src="http://server/core/img/logo-mail.gif" alt="ownCloud"/>
        					</td>
        				</tr>
        				<tr><td colspan="2">&nbsp;</td></tr>
        				<tr>
        					<td width="20px">&nbsp;</td>
        					<td style="font-weight:normal; font-size:0.8em; line-height:1.2em; font-family:verdana,'arial',sans;">
        						Hey there,<br><br>just letting you know that you now have an ownCloud account.<br><br>Your username: guiusr1<br>Please set the password by accessing it: <a href="http://server/settings/users/setpassword/form/761553510911067257740/guiusr1">Here</a><br><br>Cheers!					</td>
        				</tr>
        				<tr><td colspan="2">&nbsp;</td></tr>
        				<tr>
        					<td width="20px">&nbsp;</td>
[...]

PR #55, opened 2018-09-12
Fix user-keys encryption when user created with email

sharidas:
When user tries to set the password without logging
in for once and no session available, the user-keys
encryption was failing to create the private key
for the new password. This change helps to fix the
issue.


sharidas:
Fixes https://github.com/owncloud/encryption/issues/54

sharidas:
Backport to stable10 : https://github.com/owncloud/encryption/pull/55

phil-davis:
Good work getting all that CI to pass.

PR #65, opened 2018-10-26
webUI test to check decrypt on user specific keys

paurakhsharma:
## Description
This PR adds test for checking user-keys descript feature

Related issue: https://github.com/owncloud/enterprise/issues/2724

## How Has This Been Tested?
Locally

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Acceptance test

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Issue #33, opened 2018-05-22
missing -y option in  occ encryption:encrypt-all -y


cdamken:
### Expected behavior:

To create a script to encrypt automatically:

``` 
#!/bin bash
occ app:enable encryption
occ encryption:enable
occ encryption:select-encryption-type masterkey -y
occ encryption:encrypt-all -y
```

### Actual behavior:

Manually works except for the last command:

```
# occ app:enable encryption
encryption enabled
# occ encryption:enable
Encryption enabled

Default module: OC_DEFAULT_MODULE
# occ encryption:select-encryption-type masterkey -y
Master key successfully enabled.
occ encryption:encrypt-all -y

                                   
  The "-y" option does not exist.  
                                   

encryption:encrypt-all
```

This -y option is possible with other commands, for example:
```
occ encryption:select-encryption-type masterkey -y
```

it would be good to have the option `-y` :
```
occ encryption:encrypt-all -y
``` 

it makes no sense to have the options ` -n ` and  `-q ` if there is no ` -y ` 
```
# occ encryption:encrypt-all --help
Usage:
  encryption:encrypt-all

Options:
  -h, --help            Display this help message
  -q, --quiet           Do not output any message
  -V, --version         Display this application version
      --ansi            Force ANSI output
      --no-ansi         Disable ANSI output
  -n, --no-interaction  Do not ask any interactive question
      --no-warnings     Skip global warnings, show command output only
  -v|vv|vvv, --verbose  Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug

Help:
  This will encrypt all files for all users. Please make sure that no user access his files during this process!
```

PVince81:
Any reason why an admin would want to do this ?

Is this to automatically migrate multiple instances to a new setup ?

paurakhsharma:
@PVince81, @cdamken, @voroyam Actually if we do`encrypt-all -n` it just says
 ``You are about to encrypt all files stored in your ownCloud installation.
Depending on the number of available files, and their size, this may take quite some time.
Please ensure that no user accesses their files during this time!
Note: The encryption module you use determines which files get encrypted.``

And just abort the command, and does nothing.
What is the command supposed to do?

PVince81:
@sharidas ^

xoxys:
This is also needed for deployment automation like ansible.

pako81:
PR submitted here owncloud/core#35575 - waiting for review/approval

[...]

PR #66, opened 2018-10-31
Add skip on failing test and add web ui suite on drone

paurakhsharma:
Add skip on test scenario which is failing because of the issue below.
 Related issue: #33
Add the missing drone setup for PR: #65

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -142,7 +142,25 @@ pipeline:
       - su-exec www-data ./run.sh --remote --config /var/www/owncloud/apps/encryption/tests/acceptance/config/behat.yml --type cli
     when:
       matrix:
-        TEST_SOURCE: encryption
+        TEST_SOURCE: cli-encryption
+
+  webui-acceptance-tests:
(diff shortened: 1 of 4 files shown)


.drone.yml
```diff
@@ -142,7 +142,25 @@ pipeline:
...
+      - TEST_SERVER_URL=http://server
+      - PLATFORM=Linux
+      - BEHAT_SUITE=${BEHAT_SUITE}
+      - MAILHOG_HOST=email
+    commands:
+      - cd /var/www/owncloud/tests/acceptance/
+      - su-exec www-data ./run.sh --remote --config /var/www/owncloud/apps/password_policy/tests/acceptance/config/behat.yml --type webUI
+    when:
+      matrix:
+        TEST_SUITE: webui-acceptance
```
phil-davis: This has TEST_SUITE but the matrix entry uses
```
      TEST_SUITE: webui
      TEST_SOURCE: webui-acceptance
```
the whole file needs to be reviewed to see why it has TEST_SOURCE instead of TEST_SUITE in each place.

Issue #67, opened 2018-11-02
Move remaining CI from Travis to drone


phil-davis:

individual-it:
after #68 merged the only thing I can see that is left is the `app:check-code` check

currently only `./occ app:check-code encryption -c deprecation` would pass other tests are failing see #70 

@dpakach please add a check that runs `./occ app:check-code encryption -c deprecation` same as in https://github.com/owncloud/encryption/blob/master/.travis.yml#L39

PR #68, opened 2018-11-02
Add more tests to drone

dpakach:
Added more tests to drone to remove the  tests from travis
related to #67


individual-it:
please also add codecov

PR #77, opened 2018-12-03
Run cli acceptance tests for pgsql and PHP 7.2

phil-davis:
``TEST_SOURCE: encryption`` was a NOOP - there is no pipeline step that matches it. The 2 matrix entries here were just doing all the setup and then exiting "happy". Actually there are no specific ``api`` acceptance tests in the encryption app.

Change these matrix entries so they run the ``cli`` acceptance tests, like the matrix entry above them.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -368,21 +368,21 @@ matrix:
       OC_VERSION: daily-master-qa
       PREPARE_ACCEPTANCE: true
       USE_SERVER: true
-      TEST_SUITE: api
+      TEST_SUITE: cli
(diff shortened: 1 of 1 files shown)


phil-davis:
e.g. https://drone.owncloud.com/owncloud/encryption/370/224 executes pipeline steps all the way down to ``fix-permissions`` then does not actually run any tests.

phil-davis:
The drone matrix entries actually run some test scenarios now:
https://drone.owncloud.com/owncloud/encryption/371/216
https://drone.owncloud.com/owncloud/encryption/371/224

Issue #31, opened 2018-05-09
php-cs-fixer


mmattel:
@DeepDiver1975 
Any plans to implement ``php-cs-fixer`` as we have in core ?

mmattel:
Any update on this ?

mmattel:
@DeepDiver1975 @patrickjahns any update on this?

phil-davis:
And implement the "standard" ``Makefile`` targets for running unit tests, acceptance tests. Move ``phpunit.xml`` to the root folder...

phil-davis:
The next stage of code check/analysis is ``phan`` - see follow-on issue #80

PR #78, opened 2018-12-03
CI script refactoring

phil-davis:
Issue #31 

- refactor ``.drone.yml`` to use pipeline step names and matrix variables similar to other apps
- add CI infrastructure so that ownCloud coding style, pahn and phstan can be run from ``make``
- move the PHP unit tests into ``tests/unit`` folder
- run the core webUI and API acceptance test in 15 parts rather than 12 (in practice there are now 15 core API suites and more than 15 webUI suites. ``apiMain`` and ``apiProvisioning-v1`` happened to get run in a single matrix job and together those took about 2 hours 30 minutes - by far the longest matrix job.) Although not essential, doing this will reduce the total elapsed time to get an encryption app CI  result.
- do the ``php-cs-fixer`` ``owncloud-coding-standard`` changes to source code
- add ``owncloud-coding-standard`` job to drone matrix


.drone.yml
```diff
@@ -188,28 +241,39 @@ pipeline:
       - su-exec www-data ./run.sh --remote --type api --tags '~@skipOnEncryption&&~@skipOnEncryptionType:${ENCRYPTION_TYPE}&&~@skip'
```
individual-it: why not using `make` here?

phil-davis: This is running core acceptance tests. The local ``make test-acceptance-*`` is designed for running local app acceptance tests. I didn't consider this a major item in this refactoring PR.

I will have a play with this as a separate issue, probably I can use the local ``make test-acceptance-*`` and pass it enough env vars to point to the core Behat config and suite.

phil-davis:
Rebased and force pushed to make drone happen again.

PR #85, opened 2018-12-18
Run all 15 parts of core webUI test suites

phil-davis:
The UI core stable10 with masterkey encryption drone matrix section was only running 12 of 15 parts of the core webUI test suites. Matrix jobs for parts 13,14,15 were missed.

I noticed this while comparing drone here with user_ldap and files_primary_s3 that also do this sort of thing.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -653,6 +653,51 @@ matrix:
       TEST_SUITE: core-web-acceptance
       PART: 12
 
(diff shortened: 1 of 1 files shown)


phil-davis:
This is currently difficult to manage with the long drone matrix section.
Maybe drone v1.* will have a happier way to split out a lot of similar jobs.

PR #90, opened 2019-01-21
Integrate HSM changes to encryption

sharidas:
Integrate HSM changes to encryption

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -38,7 +38,7 @@ use OCA\Encryption\Recovery;
 use OCA\Encryption\Session;
 use OCA\Encryption\Users\Setup;
 use OCA\Encryption\Util;
-use OCP\App;
+use OCA\Encryption\Crypto\CryptHSM;
 use OCP\AppFramework\IAppContainer;
 use OCP\Encryption\IManager;
 use OCP\IConfig;
@@ -128,10 +128,25 @@ class Application extends \OCP\AppFramework\App {
 		$container->registerService('Crypt',
 			function (IAppContainer $c) {
 				$server = $c->getServer();
-				return new Crypt($server->getLogger(),
-					$server->getUserSession(),
-					$server->getConfig(),
-					$server->getL10N($c->getAppName()));
+
+				if ($this->config->getAppValue('encryption', 'hsm.url', '') !== '') {
+					$this->config->setAppValue('crypto.engine', 'internal', 'hsm');
+				}
+
+				if ($this->config->getAppValue('crypto.engine', 'internal', '') === 'hsm') {
+					return new CryptHSM($server->getLogger(),
+						$server->getUserSession(),
+						$server->getConfig(),
+						$server->getL10N($c->getAppName()),
+						$server->getHTTPClientService(),
+						$server->getRequest(),
+						$server->getTimeFactory());
+				} else {
+					return new Crypt($server->getLogger(),
+						$server->getUserSession(),
+						$server->getConfig(),
+						$server->getL10N($c->getAppName()));
+				}
 			});
 
 		$container->registerService('Session',
--- a/lib/Command/HSMDaemon.php
+++ b/lib/Command/HSMDaemon.php
@@ -1,0 +1,119 @@
+<?php
+/**
+ * @author Jörn Friedrich Dreyer <jfd@butonic.de>
+ *
+ * @copyright Copyright (c) 2017, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Encryption\Command;
+
+use OCA\Encryption\AppInfo\Application;
+use OCA\Encryption\JWT;
+use OCA\Encryption\KeyManager;
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\Encryption\IEncryptionModule;
+use OCP\Http\Client\IClient;
+use OCP\Http\Client\IClientService;
+use OCP\IConfig;
+use OCP\ILogger;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class HSMDaemon extends Command {
+
+	/** @var IClient */
+	private $httpClient;
+	/** @var IConfig */
+	private $config;
+	/** @var  ILogger */
+	private $logger;
+	/** @var ITimeFactory */
+	private $timeFactory;
+
+	/**
+	 * @param IEncryptionModule $encryption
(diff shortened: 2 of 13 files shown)


sharidas:
- [x] Integrated the HSM patch with the encryption app
- [x] Unit test added. The tests would just improve coverage.
- [x]  Command `encryption:hsmdaemon` is added as part of the patch.
- [ ] Have not tested `encryption:hsmdaemon`.

lib/AppInfo/Application.php
```diff
@@ -128,11 +128,22 @@ public function registerServices() {
...
-			});
+				$hsmUrl = $this->config->getAppValue('encryption', 'hsm.url');
+				if (\is_string($hsmUrl) && $hsmUrl !== '') {
+					return new CryptHSM($server->getLogger(),
+						$server->getUserSession(),
+						$server->getConfig(),
+						$server->getL10N($c->getAppName()),
+						$server->getHTTPClientService(),
+						$server->getRequest(),
+						$c->query('OCP\AppFramework\Utility\ITimeFactory'));
```
PVince81: I think we have `$server->getTimeFactory()`

lib/AppInfo/Application.php
```diff
@@ -128,11 +128,22 @@ public function registerServices() {
...
+				} else {
+					return new Crypt($server->getLogger(),
+						$server->getUserSession(),
+						$server->getConfig(),
+						$server->getL10N($c->getAppName()));
+				}
+			})	;
```
PVince81: this semicolon is flying away :money_with_wings:

lib/Command/HSMDaemon.php
```diff
@@ -0,0 +1,114 @@
...
+								ILogger $logger) {
+		$this->httpClient = $httpClient->newClient();
+		$this->config = $config;
+		$this->logger = $logger;
+		parent::__construct();
+	}
+
+	// TODO add route for hsmdaemon to post current secret
+	// TODO add encrypt masterkey command / as option
+	// TODO add decrypt option
```
PVince81: after the merge consider raising tickets

@butonic assuming these are not critical to have in the first version ?

lib/Command/HSMDaemon.php
```diff
@@ -0,0 +1,114 @@
...
+		$this->addOption(
+			'decrypt',
+			null,
+			InputOption::VALUE_REQUIRED,
+			'decrypt a base64 encoded value with the hsm'
+		);
+	}
+
+	protected function execute(InputInterface $input, OutputInterface $output) {
+		$hsmUrl = $this->config->getAppValue('encryption', 'hsm.url');
```
PVince81: is the default value an empty string ? I don't remember if you have to pass a default explicitly

lib/Command/HSMDaemon.php
```diff
@@ -0,0 +1,114 @@
...
+		if (\is_string($hsmUrl) && $hsmUrl !== '') {
+			$decrypt = $input->getOption('decrypt');
+			if ($decrypt) {
+				$response = $this->httpClient->post($hsmUrl, [
+					'headers' => [
+						'Authorization' => 'Bearer ' . JWT::token([
+								'exp' => \time() + 120 // 2min for clock skew
```
PVince81: use the time factory?

lib/AppInfo/Application.php
sharidas: Fixed.

lib/AppInfo/Application.php
sharidas: Done. While checking from phpstorm, I was seeing IAppContainer. And it didn't had the time factory. So I didn't touched it. Now verified in the run time. And fixed it.

PVince81:
@sharidas can you test it ? https://github.com/owncloud/encryption/pull/90#issuecomment-456345814

would also be good to have a docs ticket that explain how to set this up with the mock HSM for testing

lib/AppInfo/Application.php
```diff
@@ -128,10 +128,21 @@ public function registerServices() {
 		$container->registerService('Crypt',
 			function (IAppContainer $c) {
 				$server = $c->getServer();
-				return new Crypt($server->getLogger(),
-					$server->getUserSession(),
-					$server->getConfig(),
-					$server->getL10N($c->getAppName()));
+				$hsmUrl = $this->config->getAppValue('encryption', 'hsm.url');
```
butonic: just do document ... this causes the hsm.url app config to be two things:
1. the url for the hsm endpoint
2. the toggle for using an hsm or the internal crypto implementation

these are two different things and I have a bad feeling about it. Maybe an additional property `crypto.engine` with `internal` and `hsm` makes more sense.

lib/Command/HSMDaemon.php
butonic: correct:
- add route for hsmdaemon to post current secret
  - kill this ... I had the idea the hsmdaemon could have a command that would set up hsm based encryption and set a random shared secret. nice to have, really. Although ... it would make deployments more secure and easier to do.
- add encrypt masterkey command / as option
  - with a hsm there are actually two ways of implementing encryption:
    1. use the hsm to decrypt file individual keys
    2. use the hsm to decrypt the master key, keep it in memory (maybe even cache it with apcu), do file encryption internally
    The current implementation does the former. But we already ran into cases where every decrypt operation had to be paid for. In those cases the idea is to protect against the key being stored physically on the same machine. And the tradeof of storing it in memory vs keeping it out of the machine is driven by cost factors. Can be done in a subsequent PR. talk to @CaptionF for details.
- add decrypt option
  - this is mainly for testing and to double check the hsm setup works correctly. maybe an encrypt command makes sense as well so that the admin can try the setup on the cli ... which goes with seperating hsm.url from crypto.engine.

butonic:
@PVince81 a detailed setup is in the hsmdaemon Readme.md and Installation.md

I would merge and deal with the todos if we get opportunity for it.

PVince81:
@sharidas please check the test failures

sharidas:
Most likely this failure https://drone.owncloud.com/owncloud/encryption/479/576 has nothing to do with my change. Because the failure happens in scenario https://github.com/owncloud/core/blob/master/tests/acceptance/features/webUISharingExternal/federationSharing.feature#L293 which has a tag `@skipOnEncryption`. Let me rebase and try my luck again.

lib/Command/HSMDaemon.php
PVince81: @sharidas please raise so we can move forward

lib/AppInfo/Application.php
PVince81: a second key would be better, yes. @sharidas please add

lib/Command/HSMDaemon.php
PVince81: then adjust https://github.com/owncloud/docs/issues/586

sharidas: Tickets created:
- [x] https://github.com/owncloud/encryption/issues/100
- [x] https://github.com/owncloud/encryption/issues/101
- [x] https://github.com/owncloud/encryption/issues/102
- [ ] Adjust https://github.com/owncloud/docs/issues/586

lib/AppInfo/Application.php
sharidas: Updated the PR, by setting https://github.com/owncloud/encryption/pull/90/files#diff-b8a11151e44bcffc222c5fddc9ab92c9R133

Have also verified that it works with and without hsm.

sharidas:
I have updated the PR based on the requests made:
- https://github.com/owncloud/encryption/pull/90#discussion_r255469729
- https://github.com/owncloud/encryption/pull/90#discussion_r255472939

PVince81:
restarted cancelled build

PVince81:
can't restart the build https://drone.owncloud.com/owncloud/encryption/494/1297 :-(

maybe rebase ?

sharidas:
> can't restart the build https://drone.owncloud.com/owncloud/encryption/494/1297 :-(
> 
> maybe rebase ?

Rebasing it.

sharidas:
Backport to stable10: https://github.com/owncloud/core/pull/34527

PR #92, opened 2019-01-22
Run local server in its own folder

phil-davis:


.drone.yml
```diff
@@ -77,9 +78,9 @@ pipeline:
     image: owncloudci/php:${PHP_VERSION}
     pull: true
     commands:
-      - chown www-data /var/www/owncloud -R
-      - chmod 777 /var/www/owncloud/tests/acceptance/filesForUpload -R
-      - chmod +x /var/www/owncloud/tests/acceptance/run.sh
+      - chown www-data /var/www/owncloud/server -R
```
individual-it: could use:
```suggestion
      - chown www-data /var/www/owncloud -R
```
here and get rid of  fix-permissions-federation-server:

phil-davis: Yes, some day. CI takes >2 hours so I don't want to touch code unless I have to!

Issue #98, opened 2019-02-08
Intermittent test failure in federation sharing


dpakach:
following scenario `/tests/acceptance/features/webUISharingExternal/federationSharing.feature:292`

```
Scenario: delete a file in a folder inside a shared folder
  Given user "user1" has uploaded file "filesForUpload/textfile.txt" to "/simple-folder/simple-empty-folder/textfile.txt"
  And user "user1" from server "LOCAL" has shared "simple-folder" with user "user1" from server "REMOTE"
  And user "user1" from server "REMOTE" has accepted the last pending share
  And user "user1" re-logs in to "%remote_server%" using the webUI
  And the user opens folder "/simple-folder (2)/simple-empty-folder" using the webUI
  When the user deletes file "textfile.txt" using the webUI
  And using server "LOCAL"
  Then as "user1" file "/simple-folder/simple-empty-folder/textfile.txt" should not exist

```
is failing intermittently in drone giving following error 
```
 Then as "user1" file "/simple-folder/simple-empty-folder/textfile.txt" should not exist                                 # FeatureContext::asFileOrFolderShouldNotExist()
      Exception: file '/simple-folder/simple-empty-folder/textfile.txt' expected to not exist (status code 200, expected 401 - 404) in /var/www/owncloud/server/tests/acceptance/features/bootstrap/WebDav.php:1022

```
Drone link - https://drone.owncloud.com/owncloud/encryption/480/814

individual-it:
looks like the problem is that the file is locked. On the fed-server end we get
`172.30.0.5 - - [15/Feb/2019:13:10:44 +0000] "DELETE /remote.php/dav/files/user1/simple-folder%20(2)/simple-empty-folder/textfile.txt HTTP/1.1" 423 808 "-" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/72.0.3626.81 Safari/537.36"`

The reason for that seems to be that the server does calculate the previews, that takes time and locks the file. On the local server we disable the previews, but not on the federated server

PR #103, opened 2019-02-15
fix Intermittent test failure in federation sharing

individual-it:
debug #98


individual-it:
closing because problem seem to be identified

PR #104, opened 2019-02-20
do not trim content of downloaded file

individual-it:
the testing app does not trim the content any more (see owncloud/testing#90), so we don't want to trim it on the receiver side


individual-it:
PR for core stable10 https://github.com/owncloud/core/pull/34570

PR #106, opened 2019-03-08
Add a flag to ingore signature check

sharidas:
Add a flag to ignore the signature check for
encrypted files.

### DIFF
--- a/lib/Crypto/Crypt.php
+++ b/lib/Crypto/Crypt.php
@@ -35,6 +35,8 @@ use OCP\IConfig;
 use OCP\IL10N;
 use OCP\ILogger;
 use OCP\IUserSession;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\EventDispatcher\GenericEvent;
 
 /**
  * Class Crypt provides the encryption implementation of the default ownCloud
@@ -77,6 +79,9 @@ class Crypt {
 	/** @var IL10N */
 	private $l;
 
+	/** @var EventDispatcherInterface */
+	private $eventDispatcher;
+
 	/** @var array */
 	private $supportedCiphersAndKeySize = [
 		'AES-256-CTR' => 32,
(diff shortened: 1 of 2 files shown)


sharidas:
- [ ] Requires core PR https://github.com/owncloud/core/pull/34718. 
If you are reading this PR, kindly go through the description here https://github.com/owncloud/core/pull/34718#issue-259452874.

sharidas:
- [x] Requires https://github.com/owncloud/core/pull/34718

Updated this PR, and below is how this PR works:
Reads data from the custom symfony event and then decides if signature check has to be done or not.
Reason for using custom symfony event to read the data:
- The file reading/writing is done in encryption wrapper. So the exception thrown in Crypt is caught at wrapper, which needs a mechanism to tell the transfer-ownership command to know which file has signature mismatch. This is done via event dispatched from wrapper. Bascially the emitter of the event is the wrapper. And there are 2 listeners for this event:
1.  transfer ownership command
2. The encryption apps Crypt file

The transfer ownership command would read the events data to figure out if the file copy needs to be done second time ( or say retry ), and then the second listener to the event in the Crypt would read and updates the variable ( if required ) and the decision to enable/disable signature check happens.

lib/Crypto/Crypt.php
```diff
@@ -451,9 +467,17 @@ protected function isValidPrivateKey($plainKey) {
 	 * @throws DecryptionFailedException
 	 */
 	public function symmetricDecryptFileContent($keyFileContents, $passPhrase, $cipher = self::DEFAULT_CIPHER, $version = 0, $position = 0) {
+		$this->eventDispatcher->addListener('files.aftersignaturemismatch', function (GenericEvent $event) {
```
PVince81: let's just use a static attribute... we already have this kind of dirty practice in the decrypt all command.

PVince81: hmm now thinking this is not possible because of the encryption module abstraction.

PVince81: idea would be if there was a way to inject generic flags or configs into the encryption module

in your first attempt you used the oc_appconfig table for that, which was a bit too global.

sharidas: The whole problem is here with the code flow 

transfer commnad <----> Encryption Wrapper <-----> Crypt 

The transfer needs to output the files which are problematic and then come back for a retry in the encryption wrapper to copy the file again and crypt should know that this time no signature check.

Even I was not fully satisfied with the event logic.

Will see what can be done here..

sharidas:
Closing this PR on behalf of https://github.com/owncloud/encryption/pull/115

PR #108, opened 2019-03-11
Some tests for core issue-34705 are failing

phil-davis:
1) cut down test suites to get just the failing core webUI tests


phil-davis:
This is not happening any more. Dropping PHP 5.6 in stable10 helped.

PR #115, opened 2019-04-03
Add encryption version fix command

sharidas:
Add a new command to fix the encryption
version. This would help the admin to
fix the issues related to encryption version
mismatch for the files.

### DIFF
--- a/lib/Command/FixEncryptedVersion.php
+++ b/lib/Command/FixEncryptedVersion.php
@@ -1,0 +1,197 @@
+<?php
+/**
+ * @author Sujith Haridasan <sharidasan@owncloud.com>
+ *
+ * @copyright Copyright (c) 2019, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Encryption\Command;
+
+use OC\Files\Cache\Cache;
+use OC\Files\View;
+use OC\HintException;
+use OCP\Files\IRootFolder;
+use OCP\IUserManager;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class FixEncryptedVersion extends Command {
+	/** @var IRootFolder  */
+	private $rootFolder;
+
+	/** @var IUserManager  */
+	private $userManager;
+
+	/** @var View  */
+	private $view;
+
+	public function __construct(IRootFolder $rootFolder, IUserManager $userManager, View $view) {
+		$this->rootFolder = $rootFolder;
+		$this->userManager = $userManager;
+		$this->view = $view;
+		parent::__construct();
+	}
+
+	protected function configure() {
+		parent::configure();
+
+		$this
+			->setName('encryption:fixencryptedversion')
+			->setDescription('Fix the encrypted version if the encrypted file(s) are not downloadable.')
+			->addArgument(
+				'user',
+				InputArgument::REQUIRED,
+				'The user id whose files needs fix'
+			);
+	}
+
+	protected function execute(InputInterface $input, OutputInterface $output) {
+		$user = $input->getArgument('user');
+
+		if ($user === null) {
+			$output->writeln("<error>No user provided.</error>\n");
+		}
+
+		if ($this->userManager->get($user) === null) {
+			$output->writeln("<error>User $user does not exist. Please provide a valid user id</error>");
+			return 1;
+		}
+		$this->walkUserFolder($user, $output);
+	}
+
+	/**
+	 * @param string $user
+	 * @param OutputInterface $output
+	 */
+	private function walkUserFolder($user, OutputInterface $output) {
+		$this->setupUserFs($user);
+		$directories = [];
+		$directories[] = '/' . $user . '/files';
+		while ($root = \array_pop($directories)) {
+			$directoryContent = $this->view->getDirectoryContent($root);
+			foreach ($directoryContent as $file) {
+				$path = $root . '/' . $file['name'];
+				if ($this->view->is_dir($path)) {
+					$directories[] = $path;
+				} else {
+					$output->writeln("Verifying the content of file $path");
+					$this->verifyFileContent($path, $output);
+				}
+			}
+		}
+	}
+
+	/**
+	 * @param string $path
+	 * @param OutputInterface $output
+	 * @param bool $ignoreCorrectEncVersionCall, setting this variable to false avoids recursion
+	 */
+	private function verifyFileContent($path, OutputInterface $output, $ignoreCorrectEncVersionCall = true) {
+		try {
+			/**
+			 * In encryption, the files are read in a block size of 8192 bytes
+			 * Read block size of 8192 and a bit more (808 bytes)
+			 * If there is any problem, the first block should throw the signature
+			 * mismatch error. Which as of now, is enough to proceed ahead to
+			 * correct the encrypted version.
+			 */
+			if ($this->view->readfilePart($path, 0, 9000) !== false) {
+				$output->writeln("<info>The file $path is: OK</info>");
+			}
+			return true;
+		} catch (HintException $e) {
+			\OC::$server->getLogger()->warning("Issue: " . $e->getMessage());
+			//If allowOnce is set to false, this becomes recursive.
+			if ($ignoreCorrectEncVersionCall === true) {
+				//Lets rectify the file by correcting encrypted version
+				$output->writeln("<info>Attempting to fix the path: $path</info>");
+				return $this->correctEncryptedVersion($path, $output);
+			}
+			return false;
+		}
+	}
+
+	/**
+	 * @param string $path
+	 * @param OutputInterface $output
+	 */
+	private function correctEncryptedVersion($path, OutputInterface $output) {
+		$fileInfo = $this->view->getFileInfo($path);
+		$fileId = $fileInfo->getId();
+		$encryptedVersion = $fileInfo->getEncryptedVersion();
+		$wrongEncryptedVersion = $encryptedVersion;
+
+		$storage = $fileInfo->getStorage();
+
+		$cache = $storage->getCache();
+		$fileCache = $cache->get($fileId);
+
+		if ($storage->instanceOfStorage('OCA\Files_Sharing\ISharedStorage')) {
+			$output->writeln("<info>The file: $path is a share. Hence kindly fix this by running the script under the owner of share</info>");
+			return true;
+		}
+
+		if ($encryptedVersion >= 0) {
+			//test by decrementing the value till 1 and if nothing works try incrementing
+			$encryptedVersion--;
+			while ($encryptedVersion > 0) {
+				$cacheInfo = ['encryptedVersion' => $encryptedVersion, 'encrypted' => $encryptedVersion];
+				$cache->put($fileCache->getPath(), $cacheInfo);
+				$output->writeln("<info>Decrement the encrypted version to $encryptedVersion</info>");
+				if ($this->verifyFileContent($path, $output, false) === true) {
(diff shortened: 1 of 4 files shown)


sharidas:
- [ ] Improve the script
- [ ] Add unit tests

sharidas:
This error https://drone.owncloud.com/owncloud/encryption/599/150 is not reproducible in my local instance. Digging into it...

sharidas:
Using the unit test this is what I wanted to do:
- Simulate the scenario that happened with one of our user(s), where the filecache table got messed up with wrong encrypted version
- Run the command
- Verify everything is ok

But the problem here I find with
#### Unit test
- Simulating and executing the command is as shown in the https://github.com/owncloud/encryption/pull/115#issuecomment-479895755
- The drone failure https://drone.owncloud.com/owncloud/encryption/599/150.  Especially `openssl_seal(): not a public key (1th member of pubkeys)`
I am thinking of skipping this test for now in the unit test.

#### Acceptance test
- As of now we don't have the end point in the tests where the db could be manipulated. Once its there we could use it to alter the values and run the command and check the files. IMHO, that would be a clean way.

lib/Command/FixEncryptedVersion.php
```diff
@@ -0,0 +1,204 @@
...
+						$output->writeln("<info>Fixed the file $path with version " . $encryptedVersion . "</info>");
+						return true;
+					}
+					$encryptedVersion--;
+				}
+			}
+
+			//So decrementing did not worked. Now lets increment. Max increment is till 25
+			$increment = 1;
+			while ($increment <= 25) {
```
PVince81: not sure if we need to go that far, this would also take more time to process for many files

I suspect that the root cause would only cause a discrepancy of +1.
maybe use 5 here just to be safe.

lib/Command/FixEncryptedVersion.php
```diff
@@ -0,0 +1,204 @@
...
+				if ($oldEncryptedVersion !== null) {
```
PVince81: why not directly assign `$oldEncryptedVersion` to `$encryptedVersion` at the beginning? unnecessary extra logic makes it more confusing and less readable

lib/Command/FixEncryptedVersion.php
```diff
@@ -0,0 +1,204 @@
...
+					$encryptedVersion = $oldEncryptedVersion + $increment;
+				} else {
+					$encryptedVersion = $encryptedVersion + $increment;
```
PVince81: please recheck, from my understanding this will go as follows:

with $encryptedVersion = 6 at start
 
cycle 1: $increment=1, $encryptedVersion=6+1=7
cycle 2: $increment=2, $encryptedVersion=7+2=9
cycle 3: $increment=3, $encryptedVersion=9+3=12
...

if it doesn't do that (which might be confirmed with unit test), then rewrite the code to make it more clear what happens to the reader

tests/unit/Command/FixEncryptedVersionTest.php
```diff
@@ -0,0 +1,173 @@
...
+
+		$this->commandTester->execute([
+			'user' => self::TEST_ENCRYPTION_VERSION_AFFECTED_USER
+		]);
+
+		$output = $this->commandTester->getDisplay();
+		$this->assertEquals("Verifying the content of file /test_enc_version_affected_user1/files/hello.txt
+Decrement the encrypted version to 1
+Increment the encrypted version to 3
+Increment the encrypted version to 4
```
PVince81: please test with further increments as I seem to have spotted a bug in the increment logic... see my other comment above

lib/Command/FixEncryptedVersion.php
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	protected function configure() {
+		parent::configure();
+
+		$this
+			->setName('encryption:fixencryptedversion')
+			->setDescription('Fix the encrypted version if the encrypted file(s) are not downloadable from the UI')
+			->addArgument(
+				'user',
```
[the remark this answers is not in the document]
sharidas: I was thinking of adding that change later to the script.

lib/Command/FixEncryptedVersion.php
```diff
@@ -0,0 +1,204 @@
...
+					$this->verifyFileContent($path, $output);
+				}
+			}
+		}
+	}
+
+	/**
+	 * @param string $path
+	 * @param OutputInterface $output
+	 * @param bool $allowOnce
```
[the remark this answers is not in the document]
sharidas: Renamed this to `$ignoreCorrectEncVersionCall`. Idea is not to call 
```
$this->correctEncryptedVersion()
```
in the `catch` portion of the code.

lib/Command/FixEncryptedVersion.php
```diff
@@ -0,0 +1,204 @@
...
+	 */
+	private function verifyFileContent($path, OutputInterface $output, $allowOnce = true) {
+		try {
+			if ($this->view->readfilePart($path, 0, 9000) !== false) {
+				$output->writeln("<info>The file $path is: OK</info>");
+			}
+			return true;
+		} catch (HintException $e) {
```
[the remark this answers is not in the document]
sharidas: I didn't find anything in the code which could help us. I do understand the concern of relying on a very generic exception like `HintException` :(

lib/Command/FixEncryptedVersion.php
```diff
@@ -0,0 +1,204 @@
...
+			\OC::$server->getLogger()->warning("Issue: " . $e->getMessage());
+			//If allowOnce is set to false, this becomes recursive.
+			if ($allowOnce === true) {
+				//Lets rectify the file by correcting encrypted version
+				return $this->correctEncryptedVersion($path, $output);
```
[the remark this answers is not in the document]
sharidas: Done

lib/Command/FixEncryptedVersion.php
```diff
@@ -0,0 +1,204 @@
...
+		$fixedFileInfo = ['filename' => '', 'encryptedVersion' => 0];
+
+		$fileInfo = $this->view->getFileInfo($path);
+		$fileId = $fileInfo->getId();
+		$encryptedVersion = $fileInfo->getEncryptedVersion();
+		$oldEncryptedVersion = $encryptedVersion;
+
+		$storage = $fileInfo->getStorage();
+
+		$cache = new Cache($storage);
```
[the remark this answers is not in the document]
sharidas: Done. Thanks for the pointer.

lib/Command/FixEncryptedVersion.php
```diff
@@ -0,0 +1,204 @@
...
+		$fileCache = $cache->get($fileId);
+		$fileCacheStorage = $fileCache->getData()['storage'];
```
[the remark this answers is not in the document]
sharidas: No more relying on the string. Because in the updated version of pr I have used :
```
if ($storage->instanceOfStorage('OCA\Files_Sharing\ISharedStorage'))
```

lib/Command/FixEncryptedVersion.php
```diff
@@ -0,0 +1,204 @@
...
+
+		if (\strpos($fileCacheStorage, "shared::") !== false) {
```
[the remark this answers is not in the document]
sharidas: Done. Reference: https://github.com/owncloud/encryption/pull/115#discussion_r272562862

lib/Command/FixEncryptedVersion.php
```diff
@@ -0,0 +1,204 @@
...
+			$output->writeln("<info>The file: $path is a share. Hence kindly fix this by running the script under the owner of share</info>");
+			return true;
+		}
+
+		if ($encryptedVersion >= 0) {
+			if ($encryptedVersion === 0) {
+				$skipDecrymentedVersion = true;
+			}
+			//test by decrementing the value till 1 and if nothing works try incrementing
```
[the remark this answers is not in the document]
sharidas: Yah.

lib/Command/FixEncryptedVersion.php
sharidas: Done. Fixed with 5.

lib/Command/FixEncryptedVersion.php
sharidas: Removed this confusing logic. And made it look simple.

lib/Command/FixEncryptedVersion.php
```diff
@@ -0,0 +1,204 @@
...
+			if (($skipDecrymentedVersion !== true) && ($fixedFileInfo['filename'] !== $path)) {
+				$encryptedVersion--;
+				while ($encryptedVersion > 0) {
```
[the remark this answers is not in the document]
sharidas: Done. Removed `skipDecrymentedVersion`.

lib/Command/FixEncryptedVersion.php
```diff
@@ -0,0 +1,204 @@
...
+					if ($fixedFileInfo['filename'] === $path && $fixedFileInfo['encryptedVersion'] > 0) {
```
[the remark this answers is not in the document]
sharidas: Removed the `$filedFileInfo` usage. So no need to look into the filename and version, its already available in the loop.

lib/Command/FixEncryptedVersion.php
sharidas: Actually it doesn't work like that, I have added php comment. Also adjusted the variable `encryptedVersion` in this section to `$newEncryptedVersion`, the `oldEncryptedVersion` is renamed to `$wrongEncryptedVersion`.
So basically its like this:

when the $wrongEncryptedVersion = 6 at start

cycle 1: $increment=1, $newEncryptedVersion = 6+1 = 7
cycle 2: $increment=1, $newEncryptedVersion = 6+2 = 8
cycle 3: $increment=1, $newEncryptedVersion = 6+3 = 9
Basically the `$wrongEncryptedVersion` remains same and the `$increment` is doing the increment.

tests/unit/Command/FixEncryptedVersionTest.php
sharidas: Added further increments in the same test.

lib/Command/FixEncryptedVersion.php
PVince81: so it was only a code readability issue, cool

sharidas: Yup.

tests/unit/Command/FixEncryptedVersionTest.php
```diff
@@ -0,0 +1,201 @@
...
+	/** @var CommandTester */
+	private $commandTester;
+
+	public static function setUpBeforeClass() {
+		parent::setUpBeforeClass();
+		\OC::$server->getUserManager()->createUser(self::TEST_ENCRYPTION_VERSION_AFFECTED_USER, 'foo');
+		//Enable encryption app
+		\OC_App::enable("encryption");
+		//Enable encryption
+		\OC::$server->getConfig()->setAppValue('core', 'encryption_enabled', 'yes');
```
PVince81: normally you should save appconfig values in a variable and set that variable back at the end of the test to avoid side effects

tests/unit/Command/FixEncryptedVersionTest.php
```diff
@@ -0,0 +1,201 @@
...
+		$this->fixEncryptedVersion = new FixEncryptedVersion($this->rootFolder, $this->userManager, $this->view);
+		$this->commandTester = new CommandTester($this->fixEncryptedVersion);
+	}
+
+	/**
+	 * In this test the encrypted version is set to zero whereas it should have been
+	 * set to a positive non zero number.
+	 */
+	public function testEncryptedVersionZero() {
+		$this->markTestSkipped("Skipping this test because the user login would fail in the CI. And once we have a db endpoints available in acceptance test, we would add the test over there");
```
PVince81: please continue researching a bit

you should use `$this->loginAsUser()` for a more accurate login simulation. see https://github.com/owncloud/core/blob/v10.1.1/tests/lib/TestCase.php#L368

sharidas:
The output buffer issue https://drone.owncloud.com/owncloud/encryption/609/150 :(

sharidas:
Here is the output captured from my instance 
[a.txt](https://github.com/owncloud/encryption/files/3054520/a.txt) the output looks clean. @PVince81 let me know if this looks ok.

PVince81:
> Here is the output captured from my instance

not sure what this is and what I should do about it ?

sharidas:
> > Here is the output captured from my instance
> 
> not sure what this is and what I should do about it ?



> > Here is the output captured from my instance
> 
> not sure what this is and what I should do about it ?

Ok :) Lets wait for the CI result.

sharidas:
Will have to restart the CI again, because of unrelated failure https://drone.owncloud.com/owncloud/encryption/612/1494 :(

sharidas:
The CI has passed. Would require re-review.

PVince81:
looks good now.

I'm missing a test where the correct value is lower than the encrypted version value and where decrementing would succeed.

sharidas:
> I'm missing a test where the correct value is lower than the encrypted version value and where decrementing would succeed.

Ouch. Added the test at the end of the test file, which shows that the command would allow the files encrypted version reach the correct value by decrementing only.

Issue #118, opened 2019-04-12
Check code alignment


PVince81:
Make sure that the code on core's stable10 is the same as here on master.

Prerequisite for Github release https://github.com/owncloud/encryption/issues/17

@sharidas can you diff and align ?

sharidas:
Below is the diff I have collected based on the following criteria:
- [x] Diff is master v/s stable10 apps/encryption folder
- [x] Ignore files with **only** copyright year diff
- [x] Ignore info.xml

The diff of the files are as shown below
master v/s stable10

Excluding files with Copyright year diff and info.xml

#### appinfo/Migrations/Version20170913113840.php 

```
5c5
<  * @copyright Copyright (c) 2017, ownCloud GmbH
---
>  * @copyright Copyright (c) 2018, ownCloud GmbH
35c35
<               if (\version_compare('10.0.3', $installedVersion, '>=') === true) {
---
>               if (version_compare('10.0.3', $installedVersion, '>=') === true) {
47a48,49
> 
>
```

js/settings-admin.js
```
106,107d105
<                                       $("#encryptionRecoveryPassword").val("");
<                                       $("#repeatEncryptionRecoveryPassword").val("");
```

#### lib/Command/MigrateKeys.php
```
6c6
<  * @copyright Copyright (c) 2017, ownCloud GmbH
---
>  * @copyright Copyright (c) 2018, ownCloud GmbH
61a62
> 
100,101c101,102
<                       foreach ($this->userManager->getBackends() as $backend) {
<                               $name = \get_class($backend);
---
>                       foreach($this->userManager->getBackends() as $backend) {
>                               $name = get_class($backend);
118c119
<                               } while (\count($users) >= $limit);
---
>                               } while(count($users) >= $limit);
122a124
>
```

#### lib/Command/RecreateMasterKey.php
```
5c5
< * @copyright Copyright (c) 2017, ownCloud GmbH
---
> * @copyright Copyright (c) 2018, ownCloud GmbH
21a22
> 
130a132
> 
172a175
> 
182a186
> 
198a203
> 
240c245
<               $this->decryptAll->decryptAll($input, $output);
---
>               $this->decryptAll->decryptAll($input,$output);
258a264
>
```

#### lib/Command/SelectEncryptionType.php
```
5c5
< * @copyright Copyright (c) 2017, ownCloud GmbH
---
> * @copyright Copyright (c) 2018, ownCloud GmbH
21a22
> 
52a54
> 
81c83
[...]

PR #119, opened 2019-04-12
Remove unused use statements

phil-davis:
cleanup obviously dead code


phil-davis:
This one found more unused `use` to delete than in core `stable10` https://github.com/owncloud/core/pull/35013 - so I guess there are some differences between the 2 repos that will be looked into in issue #118

PR #120, opened 2019-04-17
Fixing encrypted version test

sharidas:
Its better to verify if the command output contains specific
strings, instead of grabbing the whole command output and
comparing with string equal.


sharidas:
Motivation behind this PR:
- Instead of grabing the whole output of the command, it would be better to check if the data that we are looking for is found in the output. Sometimes the core tar ball may have pdf and images in the skeleton and sometimes it may not. In such cases its better to go ahead with this approach.

phil-davis:
"backport" PR to core `stable10` is https://github.com/owncloud/core/pull/35039

PR #140, opened 2019-08-15
Test against core 10.3.0alpha

phil-davis:
see if this works

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1647,6 +1647,18 @@ matrix:
       DB_HOST: mysql
       USE_RELEASE_TARBALL: true
 
+    - PHP_VERSION: 7.2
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -20,6 +20,21 @@ pipeline:
       matrix:
         NEED_CORE: true
 
+  install-from-url:
+    image: owncloudci/core
+    download_url: ${DOWNLOAD_URL}
+    core_path: /var/www/owncloud/server
+    pull: true
+    exclude: "apps/encryption"
```
phil-davis: Note: the encryption app is bundled in the tarball already. So we could also do test runs using the bundled encryption code, rather than replacing with the "current master" encryption code.

phil-davis:
This is old - see PR #148

PR #148, opened 2019-10-03
Test with core latest release and 10.3.0RC1

phil-davis:
We can implement this now (like done in other app repos).
Then the drone starlark conversion can happen.


phil-davis:
CI runs for 4 hours and the job gets killed when it is nearly finished.
PR #149 splits the acceptance tests into more pieces.

PR #149, opened 2019-10-03
Split core acceptance tests into 25 parts

phil-davis:
because some pairs of acceptance test suites are taking a long time to run.

e.g. `apiShareManagementBasic` and `apiShareOperations` were running in the same job and taking close to 4 hours to finish.

(I will review all this again as I do the drone starlark conversion, but for now let's get the elapsed test time down)

- test with core latest release and 10.3.0RC1


phil-davis:
CI took 2.5 hours elapsed - better than 4 hours.

PR #156, opened 2019-11-04
Run core acceptance tests against core latest release nightly

phil-davis:
QA issue https://github.com/owncloud/QA/issues/625

1) Add code in `.drone.starlark` to allow specifying a trigger `cron` `nightly` instead of triggering on each PR.
2) Add config to run each of the acceptance test combinations against `'servers':['latest']` with `'cron':'nightly'` - this should make those run only in the nightly job.

This will add a lot of pipelines (100 extra) to the nightly test run :)

### DIFF
--- a/.drone.starlark
+++ b/.drone.starlark
@@ -158,6 +158,38 @@ config = {
 				}
 			],
 		},
+		'api-core-latest-masterkey': {
+			'suites': [
+				'apiCoreMKey',
+			],
+			'databases': [
+				'mysql:5.7',
+			],
+			'servers': [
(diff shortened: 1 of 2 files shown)


phil-davis:
CI fails:
https://drone.owncloud.com/owncloud/encryption/978/35/14

because of issue #157 

This PR is blocked.

phil-davis:
The extra pipelines did not run in the drone CI  ffor this PR, which is what is expected.
There is no way to know if they will actually run in the cron nightly until this PR is merged.

PR #176, opened 2020-02-07
Increase numberOfParts for API and webUI due to core suite splits

phil-davis:
There are more core test suites now, so have more pipelines to run them.
This helps reduce the elapsed time.


phil-davis:
This reached the limit of 1 million bytes in a generated drone configuration.
PR to make the limit higher: https://github.com/drone/drone-convert-starlark/pull/5

phil-davis:
Need to do this some other week, when we can have a bigger generated set of drone pipelines.

PR #177, opened 2020-02-07
Run core cli acceptance tests with encryption

phil-davis:
We are getting core test scenarios for CLI `occ` commands that check a user's filesystem and do other file-related things. It would be useful to know that they work with encryption.

Add drone pipelines to run them, like we have already done for core API and webUI acceptance tests.

I noticed this because of https://github.com/owncloud/user_ldap/issues/509 - in `user_ldap` we are already running some core CLI test scenarios with both user_ldap and encryption together.


phil-davis:
This got done already in #179

PR #179, opened 2020-02-10
Run core CLI acceptance tests in drone

phil-davis:
And reduce the number of parts that the core API tests are split into.
Because there is currently a limit of 1,000,000 bytes on the size of .drone.yml

Beacuse of issue https://github.com/owncloud/user_ldap/issues/509

( `user_ldap` was already running some core CLI acceptance tests with both LDAP and encryption together, so be had better run core CLI acceptance tests with encryption here)

### DIFF
--- a/.drone.starlark
+++ b/.drone.starlark
@@ -137,7 +137,7 @@ config = {
 			],
 			'runCoreTests': True,
 			'runAllSuites': True,
-			'numberOfParts': 32,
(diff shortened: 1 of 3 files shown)


phil-davis:
This should pass now. The appropriate `skipOnEncryption` has been done in core, and issues raised for problems found, e.g. #181 #182

PR #189, opened 2020-04-20
[Tests-Only] Update phpstan to latest version

dpakach:
Related - owncloud/QA#642

Update phpstan to 0.12

### DIFF
--- a/lib/Controller/RecoveryController.php
+++ b/lib/Controller/RecoveryController.php
@@ -89,12 +89,12 @@ class RecoveryController extends Controller {
 				Http::STATUS_BAD_REQUEST);
 		}
 
-		if (isset($adminEnableRecovery) && $adminEnableRecovery === '1') {
+		if ($adminEnableRecovery === '1') {
 			if ($this->recovery->enableAdminRecovery($recoveryPassword)) {
 				return new DataResponse(['data' => ['message' => (string)$this->l->t('Recovery key successfully enabled')]]);
 			}
 			return new DataResponse(['data' => ['message' => (string)$this->l->t('Could not enable recovery key. Please check your recovery key password!')]], Http::STATUS_BAD_REQUEST);
-		} elseif (isset($adminEnableRecovery) && $adminEnableRecovery === '0') {
+		} elseif ($adminEnableRecovery === '0') {
 			if ($this->recovery->disableAdminRecovery($recoveryPassword)) {
 				return new DataResponse(['data' => ['message' => (string)$this->l->t('Recovery key successfully disabled')]]);
 			}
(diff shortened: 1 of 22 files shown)


dpakach:
Some checks are failing still 
```
 ------ -------------------------------------------------------------------------------------- 
  Line   lib/Crypto/Crypt.php                                                                  
 ------ -------------------------------------------------------------------------------------- 
  119    Parameter #3 $passphrase of function openssl_pkey_export expects string, null given.  
  265    Comparison operation "<" between 269488207 and 268439583 is always false.             
 ------ -------------------------------------------------------------------------------------- 

 ------ ------------------------------------------ 
  Line   lib/Crypto/DecryptAll.php                 
 ------ ------------------------------------------ 
  208    Variable $password might not be defined.  

```

phil-davis:
PR #195 gets just the version bump done in `vendor-bin/phpstan/composer.json` without enabling it in CI.
Then we can work on this PR later to get it passing in CI.

lib/Controller/RecoveryController.php
```diff
@@ -89,12 +89,12 @@ public function adminRecovery($recoveryPassword, $confirmPassword, $adminEnableR
 				Http::STATUS_BAD_REQUEST);
 		}
 
-		if (isset($adminEnableRecovery) && $adminEnableRecovery === '1') {
+		if ($adminEnableRecovery !== null && $adminEnableRecovery === '1') {
```
phil-davis, line 92:
```suggestion
		if ($adminEnableRecovery === '1') {
```
The way it is now looks a bit silly - if `$adminEnableRecovery === '1'` then it is sure to be not null.

lib/Controller/RecoveryController.php
```diff
@@ -89,12 +89,12 @@ public function adminRecovery($recoveryPassword, $confirmPassword, $adminEnableR
...
 			if ($this->recovery->enableAdminRecovery($recoveryPassword)) {
 				return new DataResponse(['data' => ['message' => (string)$this->l->t('Recovery key successfully enabled')]]);
 			}
 			return new DataResponse(['data' => ['message' => (string)$this->l->t('Could not enable recovery key. Please check your recovery key password!')]], Http::STATUS_BAD_REQUEST);
-		} elseif (isset($adminEnableRecovery) && $adminEnableRecovery === '0') {
+		} elseif ($adminEnableRecovery !== null && $adminEnableRecovery === '0') {
```
phil-davis, line 97:
```suggestion
		} elseif ($adminEnableRecovery === '0') {
```
The way it is now looks a bit silly - if `$adminEnableRecovery === '1'` then it is sure to be not null.

Issue #193, opened 2020-04-29
Unit test errors with PHP 7.4


phil-davis:
https://drone.owncloud.com/owncloud/encryption/1233/10/4
```
There were 4 errors:

1) OCA\Encryption\Tests\Hooks\UserHooksTest::testSetPassphrase
Trying to access array offset on value of type null

/var/www/owncloud/server/apps/encryption/lib/Hooks/UserHooks.php:290
/var/www/owncloud/server/apps/encryption/tests/unit/Hooks/UserHooksTest.php:238

2) OCA\Encryption\Tests\Hooks\UserHooksTest::testSetPassphraseWithoutSession
Trying to access array offset on value of type null

/var/www/owncloud/server/apps/encryption/lib/Hooks/UserHooks.php:290
/var/www/owncloud/server/apps/encryption/tests/unit/Hooks/UserHooksTest.php:306

3) OCA\Encryption\Tests\Hooks\UserHooksTest::testSetPassphraseWithoutSessionLoggerError
Trying to access array offset on value of type null

/var/www/owncloud/server/apps/encryption/lib/Hooks/UserHooks.php:290
/var/www/owncloud/server/apps/encryption/tests/unit/Hooks/UserHooksTest.php:342

4) OCA\Encryption\Tests\Hooks\UserHooksTest::testSetPasswordNoUser
Trying to access array offset on value of type null

/var/www/owncloud/server/apps/encryption/lib/Hooks/UserHooks.php:290
/var/www/owncloud/server/apps/encryption/tests/unit/Hooks/UserHooksTest.php:378

ERRORS!
Tests: 251, Assertions: 707, Errors: 4.
```

phil-davis:
https://github.com/owncloud/encryption/blob/master/tests/unit/Hooks/UserHooksTest.php#L189
```
	public function testSetPassphrase() {
		$this->sessionMock->expects($this->exactly(4))
			->method('getPrivateKey')
			->willReturnOnConsecutiveCalls(true, false);
...
```
It gets called 4 times (when `setPassphrase` is called each of 4 times in the test), that is fine. But why are only 2 values specified in `willReturnOnConsecutiveCalls`

And `testSetPassphrase` does a whole lot of stuff setting up to supposedly `// Test first if statement` and `// Test Second if conditional` and `// Test third and final if condition` - it is not so obvious exactly which code blocks in `UserHooks.php` `setPassphrase` are intended to be executed at each point, and which ones are actually getting executed.

[...]

PR #192, opened 2020-04-29
Adjust drone pipelines for PHP 7.4

phil-davis:

PR #194, opened 2020-04-30
[Tests-Only] Return null if keyPair is null

kiranparajuli589:
##### Linked Issue:
Fixes https://github.com/owncloud/encryption/issues/193

##### Description
Return null if keyPair inside `setPassphrase` method from `UserHooks`

### DIFF
--- a/lib/Hooks/UserHooks.php
+++ b/lib/Hooks/UserHooks.php
@@ -287,6 +287,7 @@ class UserHooks implements IHook {
 				$keyPair = $this->crypt->createKeyPair("oc:uid:$user");
 
 				if ($keyPair === null) {
+					$this->logger->error('Encryption Could not update users encryption password');
 					return null;
 				}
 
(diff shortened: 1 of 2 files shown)


lib/Hooks/UserHooks.php
```diff
@@ -286,6 +286,10 @@ public function setPassphrase($params) {
 
 				$keyPair = $this->crypt->createKeyPair("oc:uid:$user");
 
+				if ($keyPair === null) {
+					return null;
```
phil-davis, line 290: Probably need to log something here (like in line 306)
And understand the reason that `createKeyPair` returns `null` in some situations.

phil-davis:
See comment https://github.com/owncloud/encryption/issues/193#issuecomment-622250916
I suspect that the problem might just be in the mocking in the unit test cases... And maybe there is no need to fix the real source code?

Because `createKeyPair` is supposed to return `array|bool` so a check of the return value should need to check `if ($keyPair === false)` anyway, rather than `if ($keyPair === null)`

Needs some advice.

phil-davis:
Finally fixed in #197 #192

PR #218, opened 2020-11-28
add path option to FixEncryptedVersion command

karakayasemi:
Part of https://github.com/owncloud/enterprise/issues/4276

- Previously, `FixEncryptedVersion` command only can be run for all the files of user. This pr adds `path` option to `FixEncryptedVersion` command to limit files to fix.
- If path indicates a file, only the file will be fixed. If indicates a directory, all the files inside directory will be fixed by command. 
- If path option exist, the user is determined by the path and the user parameter is ignored.

### DIFF
--- a/lib/Command/FixEncryptedVersion.php
+++ b/lib/Command/FixEncryptedVersion.php
@@ -57,36 +57,63 @@ class FixEncryptedVersion extends Command {
 				'user',
 				InputArgument::REQUIRED,
 				'The id of the user whose files need fixing'
+			)->addOption(
+				'path',
+				'p',
+				InputArgument::OPTIONAL,
+				'Limit files to fix with path, e.g., --path="/Music/Artist". If path indicates a directory, all the files inside directory will be fixed.'
 			);
 	}
 
 	/**
 	 * @param InputInterface $input
 	 * @param OutputInterface $output
-	 * @return int|void
+	 * @return int
 	 */
 	protected function execute(InputInterface $input, OutputInterface $output) {
 		$user = $input->getArgument('user');
+		$pathToWalk = "/$user/files";
+
+		/**
+		 * trim() returns an empty string when the argument is an unset/null
+		 */
+		$pathOption = \trim($input->getOption('path'), '/');
+		if ($pathOption !== "") {
+			$pathToWalk = "$pathToWalk/$pathOption";
+		}
 
 		if ($user === null) {
 			$output->writeln("<error>No user id provided.</error>\n");
+			return 1;
 		}
 
 		if ($this->userManager->get($user) === null) {
 			$output->writeln("<error>User id $user does not exist. Please provide a valid user id</error>");
 			return 1;
 		}
-		$this->walkUserFolder($user, $output);
+		return $this->walkPathOfUser($user, $pathToWalk, $output);
 	}
 
 	/**
 	 * @param string $user
+	 * @param string $path
 	 * @param OutputInterface $output
+	 * @return int 0 for success, 1 for error
 	 */
-	private function walkUserFolder($user, OutputInterface $output) {
+	private function walkPathOfUser($user, $path, OutputInterface $output) {
 		$this->setupUserFs($user);
+		if (!$this->view->file_exists($path)) {
+			$output->writeln("<error>Path $path does not exist. Please provide a valid path.</error>");
+			return 1;
+		}
+
+		if ($this->view->is_file($path)) {
+			$output->writeln("Verifying the content of file $path");
+			$this->verifyFileContent($path, $output);
+			return 0;
+		}
 		$directories = [];
-		$directories[] = '/' . $user . '/files';
+		$directories[] = $path;
 		while ($root = \array_pop($directories)) {
 			$directoryContent = $this->view->getDirectoryContent($root);
 			foreach ($directoryContent as $file) {
@@ -99,6 +126,7 @@ class FixEncryptedVersion extends Command {
 				}
 			}
 		}
+		return 0;
 	}
 
 	/**
@@ -134,6 +162,7 @@ class FixEncryptedVersion extends Command {
 	/**
 	 * @param string $path
 	 * @param OutputInterface $output
+	 * @return bool
 	 */
 	private function correctEncryptedVersion($path, OutputInterface $output) {
 		$fileInfo = $this->view->getFileInfo($path);
@@ -188,6 +217,7 @@ class FixEncryptedVersion extends Command {
 				$increment++;
 			}
 		}
+		return false;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


jvillafanez:
Some questions here:
* What are the consequences of skipping some "broken" files? There could be some files that I'm not aware of which might need to be fixed.
* Do we want to fix only the visible files, or we want also to fix the versions and the ones in the trashbin? This PR opens the possibility to fix files in the version and trashbin folder, which might be ok, or might not be worthy.

If we want to keep the previous behaviour and include a filter in this PR, I think it's better to keep the user as required parameter and make the path based on the user's folder. Parameters could be user=user01 path=Music/90s.
If we want to scan other folders such as trashbin or versions, the approach in this PR seems better.

karakayasemi:
>   * What are the consequences of skipping some "broken" files? There could be some files that I'm not aware of which might need to be fixed.

If there is a concern to not miss any files, the admin can run the command for whole files of user. The command can be still used like before, we will just allow to fix limited files.

>    * Do we want to fix only the visible files, or we want also to fix the versions and the ones in the trashbin? This PR opens the possibility to fix files in the version and trashbin folder, which might be ok, or might not be worthy.

https://doc.owncloud.com/server/admin_manual/configuration/files/encryption/encryption_configuration.html#which-files-are-never-encrypted
According to documentation, the version and trashbin folder can not be encrypted using this app. Only the data in the files in `data/user/files` are encrypted. Also, the current version of the command only scans the `"/$user/files"`. So, taking full path seems redundant to me also.
Actually, my purpose here was to make the path in the log file created when downloading the broken file directly usable, because the log has full path as given in the https://github.com/owncloud/enterprise/issues/4276#issuecomment-728867487. Whatever, log can be parsed. I am more close to implement the command as you suggested. Thanks for the review.

karakayasemi:
@jvillafanez now I kept the user as required parameter and made the path based on the user's folder. Please, review again. Thank you.

lib/Command/FixEncryptedVersion.php
```diff
@@ -57,36 +57,60 @@ protected function configure() {
 				'user',
 				InputArgument::REQUIRED,
 				'The id of the user whose files need fixing'
+			)->addOption(
+				'path',
+				'p',
+				InputArgument::OPTIONAL,
+				'Limit files to fix with path, e.g., --path="/Music". If path indicates a directory, all the files inside directory will be fixed.'
```
mmattel, line 64:
```suggestion
				'Limit files to fix with path, e.g., --path="Music/Artist". If path indicates a directory and/or mountpoint/directory, all the files inside directory will be fixed.'
```
This is quite nice, because you also can restrict to a mountpoint too which behaves as a directory. We have the same for the files:scan command...
I have also removed the leading "/" as this a bit confusing as mandatory element (even it gets removed later in the code). I also will drop a note regarding docs in a separate comment

mmattel:
@karakayasemi this is docs relevant. pls file a issue there and link it when merged.

lib/Command/FixEncryptedVersion.php
```diff
@@ -57,36 +57,60 @@ protected function configure() {
...
 
 	/**
 	 * @param InputInterface $input
 	 * @param OutputInterface $output
-	 * @return int|void
+	 * @return int
 	 */
 	protected function execute(InputInterface $input, OutputInterface $output) {
 		$user = $input->getArgument('user');
+		$pathToWalk = '/' . $user . '/files';
```
jvillafanez, line 75: String interpolation is preferred

lib/Command/FixEncryptedVersion.php
```diff
@@ -57,36 +57,60 @@ protected function configure() {
...
+
+		$pathOption = $input->getOption('path');
+		if ($pathOption && \trim($pathOption, '/')) {
```
jvillafanez, line 78: `--path=0` will likely be problematic here.
In addition, I'm not convinced about the `trim` inside the condition. It seems better to trim the path before the condition, which can also be reused inside the condition if needed (no need to repeat the same operation twice). I think it's easier to read too.

lib/Command/FixEncryptedVersion.php
jvillafanez: I'd prefer to keep the leading "/", such as "/Music/Artist":
* It serves as a remainder that you have to deal with it.
* It makes clearer that we're asking for a full path.

As for the mountpoint, I think it's implicit since there is no reason not to allow going inside the mountpoint. I'd expect a warning otherwise.
It's completely fine to include the mention in the docs, as well as provide additional examples, because you can be more explicit there, but for the command help I'd prefer to condense as much information as possible in the least space.

Anyway, that's a personal opinion. I'm Ok with any choice.

lib/Command/FixEncryptedVersion.php
karakayasemi: Resolved as suggested. `\trim()` returns empty string for `null`. So, strlen can be used for handle `path=0` case.

lib/Command/FixEncryptedVersion.php
```diff
@@ -57,36 +57,60 @@ protected function configure() {
...
 	 * @param OutputInterface $output
-	 * @return int|void
+	 * @return int
 	 */
 	protected function execute(InputInterface $input, OutputInterface $output) {
 		$user = $input->getArgument('user');
+		$pathToWalk = "/$user/files";
+
+		$pathOption = \trim($input->getOption('path'), '/');
+		if (\strlen($pathOption)) {
```
jvillafanez, line 78: Just a couple of details here.
* `$pathOption = \trim($input->getOption('path'), '/');` should include a comment to clarify that the scenario where the user doesn't provide a path is already handled. Something like "trim(null) will return an empty string". This also clarify the expected result in that case because you don't normally trim null values.
* Taking into account the above, the condition seem better with `if ($pathOption !== '')`. The condition seems clearer to me. I'm not sure if the `strlen` function will count all the chars, but it seems that checking against the empty string will be faster (I know it's a micro-optimization, but it's easy to do). It also feels more natural and easier to read than "if(0)" or "if(5)"

jvillafanez:
@karakayasemi there are some complains with the test coverage. Are those easy to fix? Otherwise, I'm approving the PR.
The last comment is a minor detail, but we can leave it if there is no additional change to be made. Up to you to decide.

lib/Command/FixEncryptedVersion.php
karakayasemi: Thank you for improvements. Changed as suggested.

karakayasemi:
@jvillafanez all the review points have been resolved. Also I believe, I covered all the patch with unit tests now. Please, approve the pr if you think it is ready. Thank you.

jvillafanez (approved):

mmattel:
@karakayasemi as written above, this is docs relevant. pls file a docs issue because it is merged.

@karakayasemi @jvillafanez 
because of the merge, what is the procedure that this will be part of a upcoming oC release?

phil-davis:
> what is the procedure that this will be part of a upcoming oC release?

I created release issue #220 and put it in the Releases project. That way it will be found for scheduling.

PR #222, opened 2020-12-16
Fix the hsm decrypt command

JammingBen:
## Description
The hsm decrypt command `occ encryption:hsmdaemon --decrypt DECRYPT` ist outdated and therefore does not work. This PR fixes and also reworks the command a bit: `encryption:hsmdaemon:decrypt [--username [USERNAME]] [--keyId [KEYID]] <decrypt>`. A keyId is needed to decrypt the given string. There are 3 possibilities to get one:

* Pass the keyId via option `keyId`.
* Pass the user via option `username` if you want to use the keyId of this user. *Notice*: This only works when the encryption mode is set to user.
* Leave out all options to use the masterKey as keyId. *Notice*: This only works when the encryption mode is set to master.

## Related Issue
fixes https://github.com/owncloud/enterprise/issues/4314


micbar:
needs a review

butonic (approved):

butonic:
:+1: but needs rebase

PR #223, opened 2020-12-18
[tests-only] pin mini/mc docker and skip app-required tests

phil-davis:
1) Issue https://github.com/owncloud/core/issues/38232

2) skip any core test scenarios that are tagged `@app-required`. Those would need to be run in a special CI pipeline that installs extra non-core apps. See core PR https://github.com/owncloud/core/pull/38237 for information. This should fix CI here, by skipping some new test scenarios that need both `files_texteditor` and `richdocuments` apps installed.

PR #224, opened 2020-12-20
allow binary encoding for written files to reduce file size

### LINKED ISSUE #210: Unnecessary file size overhead
Consider passing the `OPENSSL_RAW_DATA` argument to `openssl_encrypt` and `pass true for $raw_output` to `hash_hmac()` to store the files in binary instead of base64.
Base64 provides a size overhead of around 35%. Or is there a reason to use base64?

When changing this we should not forget about backwards compatibility for existing setups.

karakayasemi:
Fixes #210, https://github.com/owncloud/core/issues/10831
Needs https://github.com/owncloud/core/pull/38249

This PR makes binary encoding default and writes the encoding info to the file metadata. After this PR, all the newly created files will be encrypted with binary encoding.
For decryption, if no metadata presents we keep continue to decrypt these files with legacy base64 encoding, if header is present we use binary encoding to decrypt those files.

### DIFF
--- a/lib/Crypto/Crypt.php
+++ b/lib/Crypto/Crypt.php
@@ -109,10 +109,7 @@ class Crypt {
 		$this->config = $config;
 		$this->l = $l;
 		$this->supportedKeyFormats = ['hash', 'password'];
-		$this->useLegacyEncoding = $this->config->getSystemValue('encryption.use_legacy_encoding', false);
-		if ($this->useLegacyEncoding !== true) {
-			$this->useLegacyEncoding = false;
-		}
+		$this->useLegacyEncoding = $this->config->getSystemValue('encryption.use_legacy_encoding', false) === true;
 	}
 
 	/**
--- a/lib/Crypto/Crypt.php
+++ b/lib/Crypto/Crypt.php
@@ -62,6 +62,18 @@ class Crypt {
 	const HEADER_START = 'HBEGIN';
 	const HEADER_END = 'HEND';
 
+	/**
+	 * @var string Encoding type has been changed to binary from base64.
+	 * Reading the old format is still supported, new files are written with binary encoding by default.
+	 */
+	const DEFAULT_ENCODING_FORMAT = 'binary';
+
+	/**
+	 * @var boolean $useLegacyEncoding
+	 * Writing file with legacy base64 encoding is still supported for testing purposes
+	 */
+	private $useLegacyEncoding;
+
 	/** @var ILogger */
 	protected $logger;
 
@@ -97,6 +109,10 @@ class Crypt {
 		$this->config = $config;
 		$this->l = $l;
 		$this->supportedKeyFormats = ['hash', 'password'];
+		$this->useLegacyEncoding = $this->config->getSystemValue('encryption.use_legacy_encoding', false);
+		if ($this->useLegacyEncoding !== true) {
+			$this->useLegacyEncoding = false;
+		}
 	}
 
 	/**
@@ -207,14 +223,14 @@ class Crypt {
 			throw new \InvalidArgumentException('key format "' . $keyFormat . '" is not supported');
 		}
 
-		$cipher = $this->getCipher();
-
 		$header = self::HEADER_START
-			. ':cipher:' . $cipher
-			. ':keyFormat:' . $keyFormat
-			. ':' . self::HEADER_END;
+			. ':cipher:' . $this->getCipher()
+			. ':keyFormat:' . $keyFormat;
 
-		return $header;
+		if ($this->useLegacyEncoding !== true) {
+			$header .= ':encoding:' . self::DEFAULT_ENCODING_FORMAT;
+		}
+		return $header . ':' . self::HEADER_END;
 	}
 
 	/**
@@ -226,10 +242,11 @@ class Crypt {
 	 * @throws EncryptionFailedException
 	 */
 	private function encrypt($plainContent, $iv, $passPhrase = '', $cipher = self::DEFAULT_CIPHER) {
+		$options = $this->useLegacyEncoding === true ? 0 : OPENSSL_RAW_DATA;
 		$encryptedContent = \openssl_encrypt($plainContent,
 			$cipher,
 			$passPhrase,
-			0,
+			$options,
 			$iv);
 
 		if (!$encryptedContent) {
@@ -402,6 +419,8 @@ class Crypt {
 			$password = $this->generatePasswordHash($password, $cipher, $uid);
 		}
 
+		$binaryEncode = isset($header['encoding']) && $header['encoding'] === self::DEFAULT_ENCODING_FORMAT;
+
 		// If we found a header we need to remove it from the key we want to decrypt
 		if (!empty($header)) {
 			$privateKey = \substr($privateKey,
@@ -413,7 +432,9 @@ class Crypt {
 			$privateKey,
 			$password,
 			$cipher,
-			0
+			0,
+			0,
+			$binaryEncode
 		);
 
 		if ($this->isValidPrivateKey($plainKey) === false) {
@@ -447,10 +468,12 @@ class Crypt {
 	 * @param string $cipher
 	 * @param int $version
 	 * @param int $position
+	 * @param boolean $binaryEncode
 	 * @return string
 	 * @throws DecryptionFailedException
+	 * @throws HintException
 	 */
-	public function symmetricDecryptFileContent($keyFileContents, $passPhrase, $cipher = self::DEFAULT_CIPHER, $version = 0, $position = 0) {
+	public function symmetricDecryptFileContent($keyFileContents, $passPhrase, $cipher = self::DEFAULT_CIPHER, $version = 0, $position = 0, $binaryEncode = false) {
 		$catFile = $this->splitMetaData($keyFileContents, $cipher);
 
 		if ($catFile['signature'] !== false) {
@@ -465,7 +488,9 @@ class Crypt {
 		return $this->decrypt($catFile['encrypted'],
 			$catFile['iv'],
 			$passPhrase,
-			$cipher);
+			$cipher,
+			$binaryEncode
+		);
 	}
 
 	/**
@@ -568,14 +593,16 @@ class Crypt {
 	 * @param string $iv
 	 * @param string $passPhrase
 	 * @param string $cipher
+	 * @param boolean $binaryEncode
 	 * @return string
 	 * @throws DecryptionFailedException
 	 */
-	private function decrypt($encryptedContent, $iv, $passPhrase = '', $cipher = self::DEFAULT_CIPHER) {
+	private function decrypt($encryptedContent, $iv, $passPhrase = '', $cipher = self::DEFAULT_CIPHER, $binaryEncode = false) {
+		$options = $binaryEncode === true ? OPENSSL_RAW_DATA : 0;
 		$plainContent = \openssl_decrypt($encryptedContent,
 			$cipher,
 			$passPhrase,
-			0,
+			$options,
 			$iv);
 
 		if ($plainContent) {
@@ -685,5 +712,12 @@ class Crypt {
 		} else {
 			throw new MultiKeyEncryptException('multikeyencryption failed ' . \openssl_error_string());
 		}
+	}
+
+	/**
+	 * @return bool
+	 */
+	public function useLegacyEncoding(): bool {
+		return $this->useLegacyEncoding;
 	}
 }
--- a/lib/Crypto/Encryption.php
+++ b/lib/Crypto/Encryption.php
@@ -94,10 +94,17 @@ class Encryption implements IEncryptionModule {
 	private $decryptAll;
 
 	/** @var int unencrypted block size if block contains signature */
-	private $unencryptedBlockSizeSigned = 6072;
+	private $unencryptedBlockSizeSigned = 8096;
 
-	/** @var int unencrypted block size */
-	private $unencryptedBlockSize = 6126;
+	/** @var int signature size */
+	private $signatureSize = 54;
+
+	/**
+	 * @var boolean $useLegacyEncoding
+	 * In write operation, it is equal to crypt->useLegacyEncoding(),
+	 * In read operation, it is false if header contains "encoding:binary" otherwise true.
+	 */
+	private $useLegacyEncoding = false;
 
 	/** @var int Current version of the file */
 	private $version = 0;
@@ -174,6 +181,11 @@ class Encryption implements IEncryptionModule {
 		$this->user = $user;
 		$this->isWriteOperation = false;
 		$this->writeCache = '';
+		$this->useLegacyEncoding = true;
+
+		if (isset($header['encoding'])) {
+			$this->useLegacyEncoding = $header['encoding'] !== Crypt::DEFAULT_ENCODING_FORMAT;
+		}
 
 		if ($this->session->decryptAllModeActivated()) {
 			$encryptedFileKey = $this->keyManager->getEncryptedFileKey($this->path);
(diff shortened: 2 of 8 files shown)


karakayasemi:
@phil-davis @micbar looks like we have common problem in web ui tests independent from this pr. Daily buils are also failing. Probably, it is related with the web ui changes of oc10. (tags, open with etc.)

phil-davis:
PR #223
is currently running CI. I am trying to get a pass! When that is merged, then other PRs can be rebased.

karakayasemi:
@phil-davis thank you for information.

phil-davis:
@micbar the repo is still set to require `codecov/patch` and `codecov/project` - those should no longer be required.

micbar:
removed the check

mmattel:
@karakayasemi How does this work with existing encrypted files?
Just thinking about that the reduction is as written in the referenced PR ~35%, I would like to see that old files get the reduction too. What about backwards compatibility for existing setups? See also my comment in the referenced PR.
BTW, needs a rebase :smile:

karakayasemi:
@mmattel I already read your comments on the issue. In this pr's current implementation, the code checks `encryption.binary_encode` system config value for binary encoding. If it is not exist, the app works like before. But, I guess this config value more proper as app config. We may move this config value to app config and add a checkbox to admin settings to be able to enable.

For the migration, as you said we can add an `occ` command.

karakayasemi:
@mmattel I thought about the migration command again. Actually, we have `encryption:decrypt-all ` and `encryption:encrypt-all` commands. We may just suggest decrypt-all the files, enable the binary encode and re-encrypy all for existing setups.

What do you think? I am open to suggestions.

mmattel:
Just reading the occ command description:
`encryption:encrypt-all encrypts all data files for all users` (no single user encryption... good)
`encryption:decrypt-all decrypts all user data files, or optionally a single user` (I can live with it)

Yes, this would work
decrypt
enable new config setting
encrypt

I would thing that it would be good to have the choice, at least to see which (old or new) method will be used. That was the reason for my thoughts of a new config setting...

karakayasemi:
Another option would be making binary encoding default and writing the encoding info to the file metadata. I guess no reason to keep base64 encoding format for new files. In this way, all the newly created files encrypted with binary encoding. 
For decryption, if no metadata presents we can keep continue to decrypt these files with legacy base64 encoding, if header is present we can use binary encoding to decrypt.

mmattel:
> making binary encoding default

as long you can still reade the old format for existing files... :smile:
(means you can have both types in parallel (old: still readable, new: rw) )
The rest is a matter of documentation.

mmattel:
> old: still readable, new: rw

Idea: as old files are still readable and new files will have the new format...
What about the idea of an occ command which rewrites existing files? Somthing like:
`occ encryption:rewrite` (in single user mode only)
To think about possible options, at least -v, or a --skip= option, or a --path= option
Such a command could be interrupted and restarted without harm!

mmattel:
It is interresting that the blocksize was set to
```
old: private $unencryptedBlockSizeSigned = 6072;
new: private $unencryptedBlockSizeSigned = 8096;
```
I just think about that on the filesystem level, blocks are usually chunks in 4k. Where is 6k coming from, because that allone is producing a overhead and misaglingment ...?

karakayasemi:
@mmattel It is really not easy to understand these calculations for me either. Still I could not exactly understand the reason of magic number 6126 also. As far as I understand, ownCloud filesystem blocksize is 8192. With current base64 encoding format, we can only use 6072 bytes of in an 8192 bytes block for storing data. Its the cause of %33 file overhead. Each signed encrypted block also stores initialization vector, signature and extra padding.

Seems like storing signature of each block takes (6126-6072) = 54 bytes and if its signed the available bytes reduces to 6072. Since I only touched encoded block, I just adjusted 6072 to 8096 for binary encoded data by multiplying it with coefficient 1.33.

mrow4a:
@karakayasemi we cannot just decrypt and encrypt all customer data.. to much risk and too heavy migration to be forced on admins. This needs to be backwards compatible, meaning if you have files binary encoded (or not), this information needs to be in the metadata of the file. IN the future, if we yet another encoding type we would be able to add.

karakayasemi:
@mrow4a the current implementation is backward compatible. I added a new metadata header  `encoding` to newly written files. We can detect binary encoded files with this header. See https://github.com/owncloud/encryption/pull/224#issuecomment-751517255

phil-davis:
I guess that when an existing file is overwritten with new content, it will be encrypted using the new binary encoding. So, over time, files will gradually become encrypted by the new binary encoding.

This could probably can be a separate task for later: there could be an `occ` command that re-encrypts files. For master-key encryption, the admin could choose to re-encrypt files of a user(s). That way, after core upgrade to this feature, the admin can decide if they want to spend the effort to re-encrypt files and get the space saving. (and they do not have to go through a complete `encryption:decrypt-all` -> ` encryption:encrypt-all` process)

For user-key encryption, there are the usual challenges of how can code running as admin decrypt the files of a user - so that would be a different case.

mmattel:
> This could probably can be a separate task for later

One note here, which may help. Core has merged a PR recently where eg, Google Drive can now use subfolders for mounting. Customers may now want to use that feature. If they do so, they need to decrypt the data first, unmount, move the data into a subfolder, remount into the folder and reenable encryption for that mount - done.

For other scenarious, I fully agree that a eg. occ command for master key encryption only would be useful to rewrite data which could be made by a separately PR

mrow4a:
for this feature, I would really see that we need good QA on this, with braining what possibly could go wrong while working with files having backwards compatibility

phil-davis:
> for this feature, I would really see that we need good QA on this, with braining what possibly could go wrong while working with files having backwards compatibility

We could have some stored samples of files that have been encrypted with the old scheme, and sort out how the CI can start-up with a user that has old-style encrypted data. Then it can test that the files will download correctly and can be overwritten...

Such QA will need to be able to be done for as long as we continue to support having old/new format encrypted files mixed. That will likely be some years...

jvillafanez:
Some things trying summing up:
* Reading MUST support unencrypted files, base64-encrypted files and binary-encrypted files.
* Writing will generate always binary-encrypted files.

My only worry in the long run is that we might think the base64-encrypted files is dead code because we can't generate the files any longer. There are a couple of options to mitigate this problem:
* Include comments to ensure we must support reading base64-encrypted for backwards compatibility, as well as pointers on how to generate those files (encryption versions 1.4.0-, for example)
* Allow writing base64-encrypted files via config option.

About upgrade process:
* For small installations, the admin can run a decrypt-all + encrypt-all commands to ensure all the files are updated.
* For large installations, I expect that the files will be progressively converted when they're written in the storage. As long as we guarantee the backwards compatibility, it doesn't matter the current state of the file.

The only thing that I think it might be missing is an option for the encrypt / decrypt to perform only inside a particular path. Anyway, it's out of the scope and can be implemented separately.

For the automated testing, the following scenarios should be enough:
* Read a predefined unencrypted file knowing the expected content
* Read a predefined base64-encrypted file knowing the expected content
* Read a predefined binary-encrypted file knowing the expected content
* Writing a binary-encrypted file and compare it against the expected content
* Write random binary data and ensure we read the same content (write + read check, to ensure we can read what we're writing)
* Read a predefined base64-encrypted file, write it binary-encrypted and read it again, ensuring the contents are the same as the original.

Other than that, I don't think we need too much testing.

> We could have some stored samples of files that have been encrypted with the old scheme, and sort out how the CI can start-up with a user that has old-style encrypted data. Then it can test that the files will download correctly and can be overwritten...

I guess that adding the base64-encrypted file directly in the storage and running an `occ files:scan` won't work due to the additional flag stored in the DB to indicate that the file is encrypted. I don't think ownCloud will set such flag properly, so for ownCloud the file will be considered an unencrypted.
Hijacking the DB and set the flag directly for the file seems the easiest solution. An alternative would be start with upgrade testing: setup OC 10.5 with encryption -> upload file -> upgrade to 10.6+ -> download file.

karakayasemi:
@phil-davis I added allowing legacy base64 encoding functionality for testing purposes. It can be activated by adding `encryption.use_legacy_encoding => true` to `config.php`. But now CI started to fail with some parse errors. I am not sure it is related with changes, but I will detaily look at it.

mmattel:
This needs in case of merging an update in `config.sample.php` in core and a issue in docs to run `config-to-docs` to add this new entry: `encryption.use_legacy_encoding => true|false`

lib/Crypto/Crypt.php
```diff
@@ -226,10 +239,11 @@ public function generateHeader($keyFormat = 'hash') {
 	 * @throws EncryptionFailedException
 	 */
 	private function encrypt($plainContent, $iv, $passPhrase = '', $cipher = self::DEFAULT_CIPHER) {
+		$options = $this->useLegacyEncoding ? 0 : OPENSSL_RAW_DATA;
```
jvillafanez, line 242: We need to have consistent checks for the variable or ensure the variable is a boolean. If the admin enters "true" as string (or any other random string), the "encoding" header will be set as binary, but the contents will be written as base64.

I don't care too much about the final decision in the above scenario (we're expecting a boolean anyway), but it mustn't break the file.

lib/Crypto/Crypt.php
```diff
@@ -97,6 +109,7 @@ public function __construct(ILogger $logger, IUserSession $userSession, IConfig
 		$this->config = $config;
 		$this->l = $l;
 		$this->supportedKeyFormats = ['hash', 'password'];
+		$this->useLegacyEncoding = $this->config->getSystemValue('encryption.use_legacy_encoding', false);
```
jvillafanez, line 112: The `useLegacyEncoding` isn't guaranteed to have a boolean value here.

mrow4a: 'true'/'false'

lib/Crypto/Encryption.php
```diff
@@ -94,10 +94,17 @@ class Encryption implements IEncryptionModule {
 	private $decryptAll;
 
 	/** @var int unencrypted block size if block contains signature */
-	private $unencryptedBlockSizeSigned = 6072;
```
mrow4a, line 97: I am still worried about this one TBH

mrow4a: @DeepDiver1975 @butonic any educated guess here why was it set to 6072 ?

mrow4a: "Legacy base64 encoding reduces unencrypted block size to 3/4"

lib/Crypto/Crypt.php
mmattel: > 'yes'/'no'

why? we have many config parameters having true|false.

mmattel: true|false

I took a look in core with already existing boolean config values. maybe a hint :
`if ($includeVersion || (bool) $systemConfig->getValue('version.hide', false) === false) {`
`\OC::$server->getConfig()->getSystemValue('show_server_hostname', false) !== true,`
`\OC::$server->getConfig()->getSystemValue('use_relative_domain_name', false) === true`

mmattel:
Is https://github.com/owncloud/core/issues/38338 also the cause of failing tests here ?

phil-davis:
> Is [owncloud/core#38338](https://github.com/owncloud/core/issues/38338) also the cause of failing tests here ?

The release https://github.com/Behat/Gherkin/releases/tag/v4.7.0 causes:

https://drone.owncloud.com/owncloud/encryption/1648/31/16 
```
[Behat\Gherkin\Exception\ParserException]
Expected Step or Examples table, but got Outline on line: 466 in file: /var/www/owncloud/testrunner/tests/acceptance/features/apiShareManagementBasicToShares/createShareToSharesFolder.feature  
```

I need to pin `Behat/Gherkin` like I did in core...

phil-davis:
https://drone.owncloud.com/owncloud/encryption/1654/5/3
`No such image: owncloudci/php:7.2`
Stupid drone errors - I will restart drone CI

mmattel:
I think CI is „lost in transition“...
@phil-davis can you re-ckick it?

phil-davis:
> I think CI is „lost in transition“...
> @phil-davis can you re-ckick it?

One of the agents "hung" (the drone server says that it went for 19 hours before being cancelled!!!)

I will look...

mmattel:
@karakayasemi do you have an update to the recent code comments from above?

lib/Crypto/Crypt.php
karakayasemi: I just set to false if it is not strictly equal to true.

lib/Crypto/Crypt.php
karakayasemi: I have added strict comparison here. Also, `useLegacyEncoding` variable now should be consistently boolean.

mmattel:
Maybe reducing line 112-115 to one like:
`$this->useLegacyEncoding = ($this->config->getSystemValue('encryption.use_legacy_encoding', false) === true) ? true : false`

lib/Crypto/Encryption.php
jvillafanez: Unless I'm missing something, both the `unencryptedBlockSizeSigned` and `signatureSize` aren't assigned anywhere outside here, and they're only used inside the `getUnencryptedBlockSize` method.

I think it's better to move these variables inside the method as local variables, and add proper explanation there.

As for my reason not to use a constant, in this case, I think it's better to have the explanation of these values in only one place (inside the method that will use them) specially if there are no plans to use them anywhere else.

jvillafanez:
@karakayasemi opinion on https://github.com/owncloud/encryption/pull/224#pullrequestreview-575180609 ?


> Maybe reducing line 112-115 to one like:

```
$this->useLegacyEncoding = $this->config->getSystemValue('encryption.use_legacy_encoding', false) === true;
```
is enough. I wouldn't make a change just for this though.

lib/Crypto/Encryption.php
karakayasemi: I guess I am finally able to find the logic of that magic numbers:

```
base64 encode loss coefficient = 3/4
oc block size`= 8192

padding for unsigned block = "xx" (2 bytes)
padding for signed block`= "xxx" (3 bytes)

iv = "00iv00"(6 byte) + iv(16 byte) = 22 bytes
sign = "00sig00"(7 byte) +  sign(64 byte) = 71 bytes

Unencrypted block size for binary encoded unsigned file = (blockSize - iv - padding) = 8192-22-2 = 8168 bytes

Unencrypted block size for binary encoded signed file = (blockSize -sign - iv - padding) = 8192-71-22-3 = 8096 bytes

Unencrypted block size for base64 encoded unsigned file = (blockSize - iv - padding)*loss_coefficient = 6126 bytes

Unencrypted block size for base64 encoded signed file = (blockSize -sign - iv - padding)*loss_coefficient = 6072 bytes
```

karakayasemi:
@mmattel @mrow4a @jvillafanez I added a detailed explanation to the unencrypted block size calculation method, also moved constants inside the method as suggested. Besides, the suggested code simplification has been applied. Please, review again. Thanks.

mmattel:
@jvillafanez ready to approve?

jvillafanez (approved):

karakayasemi:
@phil-davis @micbar should we wait for acceptance tests before merging this one?

micbar:
Yes. I think we need a QA strategy here.

@karakayasemi Dou you have an idea how we could test this?

karakayasemi:
@jvillafanez already shared a good plan for testing scenarios here https://github.com/owncloud/encryption/pull/224#issuecomment-762833949 . We also allowed to write file in legacy format after discussions to be able to write test scenarios easier. I guess, it is in the to do list of QA team.

phil-davis:
I added issue #235 to add acceptance test coverage for this.
We can do it after this PR is merged, or hold this PR and we add tests on top. Whichever you like.

micbar:
@karakayasemi @phil-davis let us hold this PR and wait for the tests.

phil-davis:
@karakayasemi we were forced to switch drone CI to use `.drone.yml` in PR #237 

@dpakach is now getting the tests sorted out in PR #236 - he has rebased that.

But you will need to rebase this PR anyway so that it uses the "new" `.drone.yml`. So you may as well rebase it now. Then we can easily cherry-pick tests onto it later today or tomorrow.

phil-davis:
If everyone is happy, IMO this PR is OK to merge.

PR #237, opened 2021-02-02
[CI] move drone config to starlark and .yml format

dpakach:
move ci config to `.drone.starlark` + `.drone.yml`


dpakach:
the acceptance test pipelines are not running with this config format. The result should have been identified as drone generates .yml file from the .star file in runtime but something is different when I generate the .drone.yml file locally. 
@phil-davis do you have any idea about this?

phil-davis:
https://drone.owncloud.com/owncloud/encryption/1668 CI from a normal PR ran lots of acceptance tests a few days ago.
I will think about what could be wrong.

phil-davis:
This will be related to the use of `ctx.build.event` in `.drone.starlark`. When we locally run:
```
drone starlark --source .drone.starlark
```
then it does not know what the the build event is.

I will think what the logic should be for running locally to generate the `yml`

phil-davis:
And the locally-generated YML ends up with stuff like:
```
- name: sync-from-cache
  pull: always
  image: minio/mc:RELEASE.2020-12-18T10-53-53Z
  commands:
  - mkdir -p results
  - mc mirror cache/cache//-${DRONE_BUILD_NUMBER} results/
  environment:
    MC_HOST_cache:
      from_secret: cache_s3_connection_url
```

The local starlark of course has no idea about `ctx.repo.slug` or `ctx.build.commit` or `${DRONE_BUILD_NUMBER}` - those are all variable at the run-time of the drone build.

mmattel:
Seems that sonarcloud is not happy...

```
ERROR: Error during SonarScanner execution
--
49 | ERROR: Parameter 'sonar.pullrequest.branch' is mandatory for a pull request analysis
50 | ERROR:
51 | ERROR: Re-run SonarScanner using the -X switch to enable full debug logging.
```

Makefile
```diff
@@ -141,6 +141,11 @@ test-acceptance-core-webui: ## Run core webUI acceptance tests
 test-acceptance-core-webui: $(acceptance_test_deps)
 	BEHAT_BIN=$(BEHAT_BIN) ../../tests/acceptance/run.sh --remote --type webui -c ../../tests/acceptance/config/behat.yml --tags '~@skipOnEncryption&&~@skipOnEncryptionType:${ENCRYPTION_TYPE}&&~@skip&&~@app-required'
 
+.PHONY:
+starlark-generate:
+starlark-generate:
```
mmattel, line 146: Is this intended that `starlark-generate` is present twice?

dpakach: it's a mistake I'll remove this

phil-davis (approved):
Note: the cache for the unit test coverage ends up in a folder like `cache/cache/encryption/1aca2a32c81002fdcc78d538fe924d567940ea79-1691`

It used to be like https://drone.owncloud.com/owncloud/encryption/1668/14/5 `cache/cache/owncloud/encryption/bb82610f5118d451d03640fdd2b5eafb21e12c95-1668`

(It had an extra `owncloud` in the path)

But really that does not matter. The results get purged after being used for SonarCloud analysis.

phil-davis:
The merge CI is drone job https://drone.owncloud.com/owncloud/encryption/1692
It is running a full set of unit tests, SonarCloud analysis and acceptance tests. 
Previously the merge CI would only run the unit tests and SonarCloud analysis: https://drone.owncloud.com/owncloud/encryption/1663

That is OK for now. It might be rather challenging to run the acceptance tests just on pull-request and nightly, but not on the merge CI.

Issue #244, opened 2021-02-20
Trashbin and local storage tests failing with encryption


phil-davis:
https://drone.owncloud.com/owncloud/encryption/1734/53/16
```
--- Failed scenarios:

    /var/www/owncloud/testrunner/tests/acceptance/features/apiTrashbin/trashbinRestore.feature:164
    /var/www/owncloud/testrunner/tests/acceptance/features/apiTrashbin/trashbinRestore.feature:165
    /var/www/owncloud/testrunner/tests/acceptance/features/apiTrashbin/trashbinRestore.feature:170
    /var/www/owncloud/testrunner/tests/acceptance/features/apiTrashbin/trashbinRestore.feature:186
```

The same test scenarios failed with user_ldap + encryption: https://drone.owncloud.com/owncloud/user_ldap/2995/212/19

These started failing last night.

The scenarios pass in core, without encryption. This core PR was merged yesterday: https://github.com/owncloud/core/pull/38375 So maybe that is causing this problem.

Let's see the nightly CI result tonight...

phil-davis:
```
  Scenario Outline: Deleting a file into external storage moves it to the trashbin and can be restored                 # /var/www/owncloud/testrunner/tests/acceptance/features/apiTrashbin/trashbinRestore.feature:145
    Given using <dav-path> DAV path                                                                                    # FeatureContext::usingOldOrNewDavPath()
    And the administrator has invoked occ command "files:scan --all"                                                   # OccContext::theAdministratorHasInvokedOccCommand()
    And user "Alice" has created folder "/local_storage/tmp"                                                           # FeatureContext::userHasCreatedFolder()
    And user "Alice" has moved file "/textfile0.txt" to "/local_storage/tmp/textfile0.txt"                             # FeatureContext::userHasMovedFile()
    And user "Alice" has deleted file "/local_storage/tmp/textfile0.txt"                                               # FeatureContext::userHasDeletedFile()
    And as "Alice" the folder with original path "/local_storage/tmp/textfile0.txt" should exist in the trashbin       # TrashbinContext::elementIsInTrashCheckingOriginalPath()
    When user "Alice" restores the folder with original path "/local_storage/tmp/textfile0.txt" using the trashbin API # TrashbinContext::elementInTrashIsRestored()
    Then the HTTP status code should be "201"                                                                          # FeatureContext::thenTheHTTPStatusCodeShouldBe()
[...]

PR #248, opened 2021-03-06
[tests-only] Fix skeleton step in tests

phil-davis:


individual-it (approved):

phil-davis:
Note: after core https://github.com/owncloud/core/pull/38473 is merged, I will force-push this to trigger CI again. That will remove the failures that are due to issue #249 

Then we can see what other failures remain...

phil-davis:
Issue #244 has been fixed - the core fix should be in the daily-master-qa tarball that is built tonight. But for now, the CI here has failed because of that (they are the only fails in the drone CI)

@micbar can you force-merge this PR please?

And then we should get green CI in the nightly run tonight.

micbar (approved):

PR #252, opened 2021-03-11
Require 1.7.0

jnweiger:
Please confirm, if core 10.7.0 is a correct minimum requirement for encryption 1.5.0


karakayasemi (approved):

phil-davis (approved):

phil-davis:
I added a commit that removes the test matrix entries against "latest" core. Those cannot pass any more, because "latest" is 10.6.0.

I created issue #253 to remind us to put that back after core 10.7.0 final release - when "latest" is 10.7.0.

mmattel:
Merging?

phil-davis:
> Merging?

I will merge it - 10.7.0 core is definitely required.

PR #260, opened 2021-04-23
Fix Bad Signature exception with HSM

IljaN:


IljaN:
Maybe this change is enough: https://github.com/owncloud/encryption/pull/260/files#diff-2aee63260dfdeb0926454e39bda6db528bc389ff3de39b450646f897ebf38509R171

lib/Crypto/CryptHSM.php
```diff
@@ -168,7 +168,7 @@ public function multiKeyDecrypt($encKeyFile, $shareKey, $privateKey) { // done w
 
 			// now decode the file.
 			// version and position are 0 because we always use fresh random data as passphrase
-			$decryptedContent = $this->symmetricDecryptFileContent($encKeyFile, $decryptedKey, self::DEFAULT_CIPHER, 0, 0, true);
+			$decryptedContent = $this->symmetricDecryptFileContent($encKeyFile, $decryptedKey, self::DEFAULT_CIPHER, 0, 0, false);
```
IljaN, line 171: We maybe only need this change.

IljaN:
Continued in https://github.com/owncloud/encryption/pull/261

PR #261, opened 2021-04-26
Use binary encoding setting for HSM

### LINKED ISSUE #269: Legacy/Binary encoding option has no effect for HSM

IljaN:
This caused issues migrating HSM instances with encryption to 10.7

Fixes #269

### DIFF
--- a/lib/Crypto/CryptHSM.php
+++ b/lib/Crypto/CryptHSM.php
@@ -168,7 +168,7 @@ class CryptHSM extends Crypt {
 
 			// now decode the file.
 			// version and position are 0 because we always use fresh random data as passphrase
-			$decryptedContent = $this->symmetricDecryptFileContent($encKeyFile, $decryptedKey, self::DEFAULT_CIPHER, 0, 0, true);
+			$decryptedContent = $this->symmetricDecryptFileContent($encKeyFile, $decryptedKey, self::DEFAULT_CIPHER, 0, 0, !$this->useLegacyEncoding());
 
 			return $decryptedContent;
 		} catch (ServerException $e) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased] - XXXX-XX-XX
 
--
+- Use legacy-encoding setting for HSM also [#269](https://github.com/owncloud/encryption/issues/269)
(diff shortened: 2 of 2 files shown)


lib/Crypto/CryptHSM.php
```diff
@@ -168,7 +168,7 @@ public function multiKeyDecrypt($encKeyFile, $shareKey, $privateKey) { // done w
 
 			// now decode the file.
 			// version and position are 0 because we always use fresh random data as passphrase
-			$decryptedContent = $this->symmetricDecryptFileContent($encKeyFile, $decryptedKey, self::DEFAULT_CIPHER, 0, 0, true);
```
IljaN, line 171: Should we connect the boolean to the `encryption.use_legacy_encoding` option? https://github.com/owncloud/core/blob/d1f0c43d61a2583fdf0f457bb07ea49de100d89a/config/config.sample.php#L1405

@butonic @karakayasemi

karakayasemi: If binary encoding is breaking HSM and if we will always use it with base64 encoding, no need to add that option.

But I guess we can try to understand why HSM is failing with binary encoding. Is it possible to try HSM in the development environment? If there is any documentation, I can try it.

IljaN: Not sure about the documentation, but from what I understood the issue might happen because the old 10.5 instance was using the legacy format (we also changed that back as it defaults to false), and additionally this value had to be patched in the code to make things work.

phil-davis:
@IljaN PR #267 has switched CI to directly use `.drone.star`

If you need to change any code for this PR, then make sure to also rebase to the latest master.

lib/Crypto/CryptHSM.php
```diff
@@ -168,7 +168,7 @@ public function multiKeyDecrypt($encKeyFile, $shareKey, $privateKey) { // done w
...
+			$decryptedContent = $this->symmetricDecryptFileContent($encKeyFile, $decryptedKey, self::DEFAULT_CIPHER, 0, 0, $this->useLegacyEncoding());
```
karakayasemi, line 171: I guess we should change it as `!$this->useLegacyEncoding()`

mmattel:
When merged, docs relevant. Pls file a docs issue.

IljaN:
@karakayasemi Reworded the commit and inverted the flag

karakayasemi (approved):
LGTM!

PR #264, opened 2021-04-29
[tests-only] Split acceptance tests into more parts

SwikritiT:


phil-davis:
drone did not allow the `.drone.yml` or `.drone.star` to be any bigger than it is (1MB) So we were stuck and could not increase the number of pipelines to reduce the CI run time.  

See PR #265 for the change that was finally made.

PR #265, opened 2021-04-30
[tests-only] Adjust accceptance test matrix

### LINKED ISSUE #185: Make more parts for acceptance tests
The acceptance tests could run in more parts in drone, so that the elapsed run-time is less.

Blocked by https://github.com/drone/drone-convert-starlark/pull/5 - we have reached the limit for the size of the generated pipelines of drone.

phil-davis:
1) Remove user-keys core test pipelines
2) Increase the number of parts for core acceptance tests so that each suite ends up running in a separate pipeline

Fixes #185 

We will routinely run the local acceptance tests with both master-key and user-keys encryption.
The core tests will be run just with master-key encryption for PRs and the nightly cron job.

It is easy to switch the core acceptance tests to user-keys encryption. We can do that in a test PR whenever there is a significant change to core or encryption code, and certainly before a core or encryption app release. See PR #266 for a demo of how to run a set of acceptance tests for user-keys encryption.


phil-davis:
The longest pipeline was about 90 minutes. Previously some pipelines were taking 150 to 180 minutes.

individual-it (approved):

PR #272, opened 2021-05-18
determine encryption format from length

IljaN:


lib/Crypto/CryptHSM.php
```diff
@@ -165,10 +165,10 @@ public function multiKeyDecrypt($encKeyFile, $shareKey, $privateKey) { // done w
 				'body' => $shareKey
 			]);
 			$decryptedKey = $response->getBody();
-
+			$binaryEncode = \strlen(\bin2hex($encKeyFile)) === 256;
```
IljaN, line 168: Do we have a constant for the key-length somewhere to use here? 
Other than that LGTM

IljaN: ...in case we decide to refactor the code to use some other key-length in the future.

karakayasemi: @IljaN I moved that magic number to a constant. AFAIK, there is not other usage of this length.

PR #273, opened 2021-05-18
Restore encrypted version to original if no fix was found

### LINKED ISSUE #275: fix-encrypted version command keeps version even if no fix was found
When trying to debug signature problems on a production instance the fix-encryption command does not restore the version to the original value even after no fix was found.

IljaN:
Fixes #275


karakayasemi (approved):
LGTM!

mmattel:
docs relevant, pls file a docs issue, thanks

Issue #315, opened 2021-12-28
encryption PHP unit tests fail


phil-davis:
https://drone.owncloud.com/owncloud/encryption/2223/4/4
```
7.4: Pulling from owncloudci/php
Digest: sha256:ccda771b8c1582336b853aceb9959e842f125d6d51a65f5ecde50a070675aaed
Status: Image is up to date for owncloudci/php:7.4
+ make test-php-unit-dbg
phpdbg -qrr -d memory_limit=4096M -d zend.enable_gc=0 "../../lib/composer/bin/phpunit" --configuration ./phpunit.xml --testsuite unit
PHPUnit 9.5.10 by Sebastian Bergmann and contributors.

Runtime:       PHPDBG 7.4.27
Configuration: ./phpunit.xml

FFF............................................................  63 / 259 ( 24%)
............................................................... 126 / 259 ( 48%)
............................................................... 189 / 259 ( 72%)
............................................................... 252 / 259 ( 97%)
.......                                                         259 / 259 (100%)

Time: 00:11.840, Memory: 554.42 MB

There were 3 failures:

1) OCA\Encryption\Tests\Command\FixEncryptedVersionTest::testEncryptedVersionIsNotZero
Failed asserting that 'Verifying the content of file /test_enc_version_affected_user1/files/hello.txt\n
The file /test_enc_version_affected_user1/files/hello.txt is: OK\n
Verifying the content of file /test_enc_version_affected_user1/files/ownCloud Manual.pdf\n
The file /test_enc_version_affected_user1/files/ownCloud Manual.pdf is: OK\n
Verifying the content of file /test_enc_version_affected_user1/files/world.txt\n
The file /test_enc_version_affected_user1/files/world.txt is: OK\n
Verifying the content of file /test_enc_version_affected_user1/files/Photos/Lake-Constance.jpg\n
The file /test_enc_version_affected_user1/files/Photos/Lake-Constance.jpg is: OK\n
Verifying the content of file /test_enc_version_affected_user1/files/Photos/Portugal.jpg\n
The file /test_enc_version_affected_user1/files/Photos/Portugal.jpg is: OK\n
Verifying the content of file /test_enc_version_affected_user1/files/Photos/Teotihuacan.jpg\n
The file /test_enc_version_affected_user1/files/Photos/Teotihuacan.jpg is: OK\n
Verifying the content of file /test_enc_version_affected_user1/files/Documents/Example.odt\n
[...]

PR #316, opened 2021-12-29
run tests with 10.9.1RC1 instead of master

individual-it:
because of https://github.com/owncloud/encryption/issues/315 some tests fail with latest, so run them with 10.9.1RC1
this PR needs to be reverted after 10.9.1 is released and updated if there is an other RC

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -55,7 +55,7 @@ config = {
             ],
             "servers": [
                 "daily-master-qa",
-                "latest",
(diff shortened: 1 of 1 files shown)


JammingBen (approved):

jnweiger:
@phil-davis revert this now?

phil-davis:
> @phil-davis revert this now?

yes, see #323

PR #318, opened 2021-12-30
run webUI core tests not on latest but on 10.9.1RC1

individual-it:
same as #316 but for a last suite
this suite only runs nightly, that is why #316 did not fail

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -306,13 +306,13 @@ config = {
         },
         "webUI-core-latest-masterkey": {
             "suites": [
-                "webUIcoreMKey",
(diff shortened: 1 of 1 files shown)


.drone.star
```diff
@@ -306,13 +306,13 @@ config = {
         },
         "webUI-core-latest-masterkey": {
             "suites": [
-                "webUIcoreMKey",
+                "webUIcoreMK",
```
individual-it, line 309: name got too long :-(

JammingBen (approved):

PR #355, opened 2022-07-20
[DO-NOT-MERGE] Disable signature check via env-variable

IljaN:


lib/Crypto/Crypt.php
```diff
@@ -496,15 +496,19 @@ protected function isValidPrivateKey($plainKey) {
...
 		$catFile = $this->splitMetaData($keyFileContents, $cipher);
 
-		if ($catFile['signature'] !== false) {
-			try {
-				$this->checkSignature($catFile['encrypted'], $passPhrase . $version . "-" . $position, $catFile['signature']);
-			} catch (HintException $e) {
-				// Check legacy format...
-				$this->checkSignature($catFile['encrypted'], $passPhrase . $version . $position, $catFile['signature']);
+
+		if (\getenv('ENCRYPTION_DISABLE_SIGNATURE', false) === false) {
```
butonic, line 500: Are you sure getenv will return a bool type false/true?

butonic (approved):

PR #360, opened 2022-07-26
test size when switching encodings

butonic:
I'm trying to find a code path that triggers the wrong recalculation of the unencrypted filesize.

In my case I have a 19686944 byte encrypted file on disk but a 19447696 byte size in the db. The difference seems to be the overhead for bunary encoded chunks.

Signed-off-by: Jörn Friedrich Dreyer <jfd@butonic.de>


ownclouders:
:boom: Acceptance tests pipeline <strong>apiMasterKeys-latest-mysql8.0-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/encryption/2584/18

butonic:
the test needs to set `encryption.use_legacy_encoding` and then
1. copy an encrypted file and the corresponding key on disk,
2. run a file scan command to pick up the encrypted file from disk (kind of mimicking a backup restore)
3. run a fix versions command
now the wrong filesize will have been written to the database: see the [top post](https://github.com/owncloud/encryption/pull/360#issue-1318035322) for calculation details.

with https://github.com/owncloud/core/pull/40240 the miscalculation is fixed

phil-davis (approved):
It is good to check the size of encrypted files as seen by the server.
We could have better steps to avoid having to use a When step like:
`user "Alice" gets the size of file "/textfile1.txt" using the WebDAV API`
in the `Then` section of the scenarios.
But those steps are already like that in core, so we can live with a bit of "non-standard" scenario step format here.

phil-davis:
I will merge this when CI finishes, then make an issue for what other test scenario(s) might be useful to have.

PR #361, opened 2022-07-27
Handle the versions in the trashbin for the checksum verify command

jvillafanez:
`files:checksum:verify` was failing due to problems finding the right key with the versions that were in the trashbin. This was caused by a wrong computation for the right path where the key is.

Rel https://github.com/owncloud/enterprise/issues/5272

### DIFF
--- a/lib/Crypto/Encryption.php
+++ b/lib/Crypto/Encryption.php
@@ -546,13 +546,16 @@ class Encryption implements IEncryptionModule {
 	 * @return string
 	 */
 	protected function getPathToRealFile($path) {
+		// By default, the "real path" is the same as the original path
 		$realPath = $path;
 		$parts = \explode('/', $path);
 		if ($parts[2] === 'files_versions') {
+			// for versions, we need to point to the actual file, not the version of the file
 			$realPath = '/' . $parts[1] . '/files/' . \implode('/', \array_slice($parts, 3));
 			$length = \strrpos($realPath, '.');
 			$realPath = \substr($realPath, 0, $length);
 		} elseif ($parts[2] === 'files_trashbin' && $parts[3] === 'versions') {
+			// if the version is in the trashbin, we need to point to the actual file inside the trashbin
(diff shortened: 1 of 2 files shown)


lib/Crypto/Encryption.php
```diff
@@ -552,6 +552,9 @@ protected function getPathToRealFile($path) {
 			$realPath = '/' . $parts[1] . '/files/' . \implode('/', \array_slice($parts, 3));
 			$length = \strrpos($realPath, '.');
 			$realPath = \substr($realPath, 0, $length);
+		} elseif ($parts[2] === 'files_trashbin' && $parts[3] === 'versions') {
+			$realPath = "/{$parts[1]}/files_trashbin/files/" . \implode('/', \array_slice($parts, 4));
+			$realPath = \preg_replace('/.v[0-9]+([^\/]*)$/', '$1', $realPath);
 		}
 
 		return $realPath;
```
mrow4a, line 560: @jvillafanez maybe I am too picky, but would not it be better if the if-else block actually checked all possible combinations and final else throws an exception on unsupported? This way we could avoid an issues like this one. Thoughts?

mrow4a (commented):
nice catch, adding small comment

lib/Crypto/Encryption.php
jvillafanez: This piece of code also runs for regular files, and for those files the expected path matches the original, so there is nothing to do. Even if we throw an exception here, we'll need to handle the key location for the regular files. I think it's better for the code to go through this function in order to handle all the cases in the same place.

pako81:
confirmed it resolves the exception for keys of versions in trashbin 👍

mrow4a (approved):

pako81:
merge or something more to discuss?

phil-davis:
I  created release issue #362 and put it in the qualification column of the Releases project so that someone can decide when to do a release.

Issue #369, opened 2022-09-20
Different scenario has failed at recent nightly build


amrita-shrestha:
Related to occ command
There were some changes recently in PR https://github.com/owncloud/core/pull/40359

Build: https://drone.owncloud.com/owncloud/encryption/2673
https://drone.owncloud.com/owncloud/encryption/2674
https://drone.owncloud.com/owncloud/encryption/2675

Failed Scenario: 
```
cliEncryption/recreatemasterkey.feature:4
cliEncryption/recreatemasterkey.feature:9
webUIMasterKeyType/webUIMasterKeys.feature:7
webUIMasterKeyType/webUIMasterKeys.feature:19
webUIUserKeysType/userKeys.feature:14
```

Failed step:
```feature
  Scenario: recreate masterkey                                                                                                    # /var/www/owncloud/testrunner/apps/encryption/tests/acceptance/features/cliEncryption/recreatemasterkey.feature:4
    Given user "admin" has uploaded file "filesForUpload/textfile.txt" to "/somefile.txt"                                         # FeatureContext::userHasUploadedAFileTo()
    When the administrator successfully recreates the encryption masterkey using the occ command                                  # EncryptionContext::recreateMasterKeyUsingOccCommand()
      The occ command did not run  (Exception)
    Then the downloaded content when downloading file "/somefile.txt" for user "admin" with range "bytes=0-6" should be "This is" # 

Scenario: recreate masterkey and upload data                                                                               # /var/www/owncloud/testrunner/apps/encryption/tests/acceptance/features/cliEncryption/recreatemasterkey.feature:9
    Given user "Alice" has been created with default attributes and small skeleton files                                     # FeatureContext::userHasBeenCreatedWithDefaultAttributes()
    And user "Alice" has uploaded file "filesForUpload/textfile.txt" to "/somefile.txt"                                      # FeatureContext::userHasUploadedAFileTo()
    When the administrator successfully recreates the encryption masterkey using the occ command                             # EncryptionContext::recreateMasterKeyUsingOccCommand()
      The occ command did not run  (Exception)
    And user "Alice" uploads chunk file "1" of "1" with "AA" to "/somefile.txt" using the WebDAV API                         # FeatureContext::userUploadsChunkedFile()
    Then the downloaded content when downloading file "/somefile.txt" for user "Alice" with range "bytes=0-3" should be "AA" # FeatureContext::downloadedContentWhenDownloadingForUserWithRangeShouldBe()

[...]

PR #370, opened 2022-09-20
[tests-only] Save exitcode of last occ command

saw-jan:
## Description
The PR https://github.com/owncloud/core/pull/40359 has implemented different method to save the exitcode of the last occ command and the local tests were failing because the exitcode was not saved.
In this PR, I have implemented the updated way to save the occ exitCode.

## Related Issue
Fixes https://github.com/owncloud/encryption/issues/369


phil-davis (approved):
LGTM if CI passes

PR #378, opened 2022-11-28
[tests-only] [full-ci] Run phan with PHP 7.3

phil-davis:
`phan` will report errors if any of the code is not compatible with PHP 7.3.

This app still supports PHP 7.3 when used with oC10 core 10.11 and earlier.

To get `phan` passing, I suppressed most of the things it reported. They all look like OK code to me, just deprecated calls, and some code idioms that phan thinks are a bit odd.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -47,6 +47,14 @@ config = {
     ],
     "codestyle": True,
     "phpstan": True,
+    "phan": {
+        "multipleVersions": {
+            "phpVersions": [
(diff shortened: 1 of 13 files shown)


phil-davis:
https://drone.owncloud.com/owncloud/encryption/2790/5/3
`phan` passed on PHP 7.3 (and 7.4) - good.

SwikritiT (approved):

PR #382, opened 2023-02-14
[docs-only] Update readme to metion that encryption is supported only on oppenssl v1

SwikritiT:
I had a hard time yesterday trying to find out why I was getting `503` with the message 
`Encryption not ready: multikeyencryption failed error:0480006C:PEM routines::no start line` and later on found out that the encryption was not supported in my ubuntu 22 as it has openssl v3. While it's mentioned here https://doc.owncloud.com/server/next/admin_manual/installation/manual_installation/manual_installation_prerequisites.html#openssl-version

I think it's relevant to mention it in the readme as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,3 +40,6 @@ With user-key encryption where only the specific user (not even the admin) can d
 `config:app:set encryption userSpecificKey --value 1/0`
 
 Set to 1 if userKey encryption is enabled
+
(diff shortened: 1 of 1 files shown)


phil-davis (commented):
Text looks OK. @mmattel please give any comments.

Issue #385, opened 2023-02-22
`phpstan` failing at nightly build


SagarGi:
### Description

`phpstan` pipeline is failing on nightly build and seems to be a hard failure. Needs some investigation regarding it.

```
Line   lib/Crypto/Encryption.php                                           
 ------ -------------------------------------------------------------------- 
  321    Comparison operation ">" between 0 and 0 is always false.           
         💡 Because the type is coming from a PHPDoc, you can turn off this   
            check by setting treatPhpDocTypesAsCertain: false in your        
            ./phpstan.neon.                                                  
 ------ -------------------------------------------------------------------- 


 [ERROR] Found 1 error                                                          

make: *** [Makefile:112: test-php-phpstan] Error 1
```

Failed Build: https://drone.owncloud.com/owncloud/encryption/2914/2/4

phil-davis:
`Installing phpstan/phpstan (1.10.1)`

https://github.com/phpstan/phpstan/releases/tag/1.10.1

1.10.0 and 1.10.1 were released in the last day. I think that they are reporting a "false positive". 

The code is:
`while (\strlen($data) > 0) {`

And it looks to me that `$data` is an ordinary string that can have a length greater than zero or equal to zero.

phil-davis:
I raised an issue with phpstan
https://github.com/phpstan/phpstan/issues/8938

phil-davis:
@SagarGi I will followup this. If the phpstan issue is not fixed quickly, then I will pin the `phpstan` version to a previous passing version. That will get CI to get past `phpstan` so that we can see if everything else passes.

SagarGi:
> @SagarGi I will followup this. If the phpstan issue is not fixed quickly, then I will pin the `phpstan` version to a previous passing version. That will get CI to get past `phpstan` so that we can see if everything else passes.

sure !!

PR #386, opened 2023-02-22
Pin phpstan to 1.9.18

phil-davis:
Related issue #385 

If we do not get any other fix quickly, we can pin phpstan to 1.9.18 (the version yesterday that passes)


phil-davis:
Let's merge this to get CI green. I just need an approval.

SagarGi (approved):
LGTM :+1:

PR #389, opened 2023-03-28
feat: drop setup of user based encryption

DeepDiver1975:


ownclouders:
:boom: Acceptance tests pipeline <strong>apiMasterKeys-latest-mysql8.0-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/encryption/2968/16

ownclouders:
:boom: Acceptance tests pipeline <strong>cliEncryption-latest-postgres9.4-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/encryption/2968/13

DeepDiver1975:
Yes. As soon as encryption is enabled and the encryption app is enabled as well master key is enabled as well.
Unless user key was already there before.

phil-davis:
https://drone.owncloud.com/owncloud/encryption/2971/6/5
```
There was 1 failure:

1) OCA\Encryption\Tests\Hooks\UserHooksTest::testLogin
Expectation failed for method name is "getAppValue" when invoked 1 time(s).
Method was expected to be called 1 times, actually called 0 times.
```

phil-davis:
Code looks good.

For support and regression testing, do we still want some way to create an oC10 install with user-key encryption?

For example, support is investigating some issue with a production installation somewhere and wants to try and reproduce the problem. How can the support team startup a test oC10, and enable encryption with user-key?

And as long as we still support user-key encryption, then we want to avoid accidentally introducing regressions. So it would be good t have some way that a test environment (preferably automated) can enable user-key encryption and then run the test suite (and/or some manual tests). 

Maybe it is as easy as:
```
\OC::$server->getAppConfig()->getValue("encryption", "userSpecificKey", "")
```
setting that app config value to 1, and "it will just happen"?

mrow4a:
@phil-davis I think user-key was deprecated in 10.7, which was very long ago. I think it is fair to stop testing it by now already.

mrow4a (approved):

mrow4a:
related https://github.com/owncloud/enterprise/issues/4939

PR #390, opened 2023-03-28
feat: php7.4 is minimum now

DeepDiver1975:


.drone.star
```diff
@@ -51,7 +51,6 @@ config = {
         "multipleVersions": {
             "phpVersions": [
                 DEFAULT_PHP_VERSION,
-                "7.3",
```
phil-davis, line 54: Actually you can remove the whole thing, and just say:
```
"phan": True,
```

Because it defaults to running just for `DEFAULT_PHP_VERSION`

Issue #409, opened 2024-03-04
Nightly: webUI test failing [acceptance]


Salipa-Gurung:
Build: 
https://drone.owncloud.com/owncloud/encryption/3462/16/15

Failed Scenario:
- tests/acceptance/features/webUIMasterKeyType/webUIMasterKeys.feature:8
- tests/acceptance/features/webUIMasterKeyType/webUIMasterKeys.feature:21

`Scenario: user can access their file after recreating master key with re-login  is passed after rerun`
```feature
Scenario: user cannot access their file after recreating master key with re-login                         # /var/www/owncloud/testrunner/apps/encryption/tests/acceptance/features/webUIMasterKeyType/webUIMasterKeys.feature:8
    Given user "Alice" has been created with default attributes and large skeleton files                    # FeatureContext::userHasBeenCreatedWithDefaultAttributes()
    And the administrator has set the encryption type to "masterkey"                                        # EncryptionContext::theAdministratorSetsEncryptionTypeToUsingTheOccCommand()
    And user "Alice" has uploaded file "filesForUpload/textfile.txt" to "/somefile.txt"                     # FeatureContext::userHasUploadedAFileTo()
    And user "Alice" has logged in using the webUI                                                          # WebUILoginContext::theUserHasLoggedInUsingTheWebUI()
    When the administrator successfully recreates the encryption masterkey using the occ command            # EncryptionContext::recreateMasterKeyUsingOccCommand()
    Then the command output should contain the text 'Note: All users are required to relogin.'              # OccContext::theCommandOutputContainsTheText()
INFORMATION: timed out waiting for ajax calls to start
INFORMATION: timed out waiting for outstanding ajax calls
    When the user opens file "lorem.txt" expecting to fail using the webUI                                  # WebUIFilesContext::theUserOpensFolderNamedUsingTheWebUI()
      │ INFORMATION: timed out waiting for ajax calls to startINFORMATION: timed out waiting for outstanding ajax calls
    Then the user should be redirected to the general exception webUI page with the title "%productname%"   # WebUIGeneralContext::theUserShouldBeRedirectedToGeneralExceptionPage()
      Page\OwncloudPage::waitTillXpathIsVisible xpath: .//span[@class='error error-wide']//h2[1] cannot find element (SensioLabs\Behat\PageObjectExtension\PageObject\Exception\ElementNotFoundException)
[...]

PR #410, opened 2024-03-06
[tests-only] [full-ci] fixing nightly 

ghost:
related issue:
https://github.com/owncloud/encryption/issues/409

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -321,28 +321,28 @@ config = {
 }
 
 def main(ctx):
-    before = beforePipelines(ctx)
+    #before = beforePipelines(ctx)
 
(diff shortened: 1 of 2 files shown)


phil-davis:
I checked the various oC10 app nightly CI pipelines. They use https://download.owncloud.com/server/daily/owncloud-daily-master-qa.tar.bz2 as the core code, and that daily tarball is up-to-date (it gets created each night from the current core master branch).

And https://download.owncloud.com/server/stable/owncloud-latest.tar.bz2 has the core code of the latest release 10.14.0. So pipelines that run with core latest are also using the expected core code.

PR #412, opened 2024-12-23
[tests-only][full-ci] skip master key test

ghost:
Master key test has been failing for a while now(which causes next test to fail). We could not figure out what caused the problem.
For now skipping test:
```
Scenario: user cannot access their file after recreating master key with re-login
```

so nightly is not red.
During debugging process, found out **core crash** is the reason behind flaky test, For more detail, visit this comment: https://github.com/owncloud/encryption/issues/409#issuecomment-2008636281

related issue: #409


amrita-shrestha (approved):
lgtm

